PR #9, opened 2025-08-30
Added context api and added missing load_dotenv()

### LINKED ISSUE #8: [ENHANCEMENT] Add context API support
The current langgraph platform supports context along with config["configurable"]. The goal of this issue is to implement the same.

[Reference](https://github.com/langchain-ai/langgraph/pull/5243)

arthiondaena:
closes #8 

Replaced `graphs/react_agent` with the langgraph's [react_agent template](https://github.com/langchain-ai/react-agent/tree/main), to show that aegra is compatible with the langgraph.

As for the missing `load_dotenv()`, the migrations(`alembic/env.py`) and also the fastapi application(`run_server.py`) were not loading the `.env` file before the migration or application start. Fixed it by adding `load_dotenv()` everywhere necessary.

Implemented [backward compatibility](https://github.com/arthiondaena/aegra/blob/4a8057f206cdeb2a8ab02fa6178919108bb7ade5/src/agent_server/api/assistants.py#L59-L69) with the existing config["configurable"]


graphs/react_agent/context.py
```diff
@@ -24,7 +22,7 @@ class Configuration:
     )
 
     model: Annotated[str, {"__template_metadata__": {"kind": "llm"}}] = field(
-        default="openai/gpt-4o-mini",
+        default="anthropic/claude-3-5-sonnet-20240620",
```
ibbybuilds: can we keep default model as gpt-4o-mini please

arthiondaena: Fixed

PR #24, opened 2025-09-10
feat: add subgraph streaming and skipping events with langsmith:nostream tag

tokezooo:
## Overview
This PR introduces streaming for subsequent graphs and skipping event logic. This is useful for supervisor/router agent architectures: subgraph events now stream properly instead of the final subgraph result being streamed in one single event. 

Events with `langsmith:nostream` tag are removed from streaming(see [Hiding messages in Agent Chat UI](https://github.com/langchain-ai/agent-chat-ui?tab=readme-ov-file#hiding-messages-in-the-chat)). This tag may be used for router node responses so they can be hidden from displaying on frontend.

## Key Changes

- Set `subgraphs=True` for `graph.astream()` called inside `src/agent_server/api/runs.py/execute_run_async()` 
- Add `_should_skip_event()` to `src/agent_server/api/runs.py`

### DIFF
--- a/src/agent_server/api/runs.py
+++ b/src/agent_server/api/runs.py
@@ -671,7 +671,7 @@ async def execute_run_async(
     interrupt_before: Optional[Union[str, List[str]]] = None,
     interrupt_after: Optional[Union[str, List[str]]] = None,
     multitask_strategy: Optional[str] = None,
-    subgraphs: Optional[bool] = None,
+    subgraphs: Optional[bool] = False,
 ):
     
     """Execute run asynchronously in background using streaming to capture all events"""    # Use provided session or get a new one
@@ -739,16 +739,13 @@ async def execute_run_async(
             
         only_interrupt_updates = not user_requested_updates
         
-        # Use streaming service's broker system to distribute events
-        # Determine subgraphs parameter: if explicitly set, use that value; otherwise use False as default
-        subgraphs_param = False if subgraphs is None else subgraphs
         
         async with with_auth_ctx(user, []):
             async for raw_event in graph.astream(
                 execution_input,
                 config=run_config,
                 context=context,
-                subgraphs=subgraphs_param,
+                subgraphs=subgraphs,
                 stream_mode=final_stream_modes,
             ):
                 # Skip events that contain langsmith:nostream tag
--- a/src/agent_server/models/runs.py
+++ b/src/agent_server/models/runs.py
@@ -45,7 +45,7 @@ class RunCreate(BaseModel):
     
     # Subgraph configuration
     stream_subgraphs: Optional[bool] = Field(
-        None,
+        False,
         description="Whether to include subgraph events in streaming. When True, includes events from all subgraphs. When False (default when None), excludes subgraph events. Defaults to False for backwards compatibility.",
     )
     


ibbybuilds:
@tokezooo Thanks! Can you also add tests for this functionality?

- e2e would be best for covering subgraph and skipping tag

- unit tests for the method too, if you have time

cclehui:
is the tag "langsmith:nostream"  need a langsmith key to work ?    is it need langsmith init ?

tokezooo:
> is the tag "langsmith:nostream" need a langsmith key to work ? is it need langsmith init ?

No, you dont need any keys or any Langsmith/Platform inits

src/agent_server/api/runs.py
```diff
@@ -709,8 +728,13 @@ def _normalize_mode(mode):
                 execution_input,
                 config=run_config,
                 context=context,
+                subgraphs=True,
```
ibbybuilds: this is too aggressive to set subgraphs=true for every run. there should be a boolean option for subgraph in the request, and we should set it to true or false based on that.

e2e/test_event_filtering_and_subgraphs.py
```diff
@@ -0,0 +1,331 @@
...
+                "event_number": event_count,
+            },
+        )
+
+        if getattr(chunk, "event", None) == "end":
+            completed = True
+            break
+
+    # Validate that streaming completed successfully
+    assert completed, "Stream should complete with an 'end' event"
```
ibbybuilds: this doesn't hold true anymore we have remoed the end event from stream now it just ends with whatever was the last event from the graph

e2e/test_event_filtering_and_subgraphs.py
```diff
@@ -0,0 +1,331 @@
...
+
+    # Note: We can't directly test that langsmith:nostream events were filtered
+    # since they would be invisible to the client. The test validates that
+    # the filtering doesn't break the streaming flow.
+    elog("Event filtering test completed", {"total_events": event_count})
+
+
+@pytest.mark.e2e
+@pytest.mark.asyncio
+async def test_subgraphs_streaming_parameter_e2e():
```
ibbybuilds, line 116: this test doesn’t actually cover subgraph streaming since the graph being used doesn’t even have a subgraph. for a proper e2e test, please add a simple subgraph agent that can be tested against.

src/agent_server/api/runs.py
```diff
@@ -718,13 +740,21 @@ def _normalize_mode(mode):
         only_interrupt_updates = not user_requested_updates
         
         # Use streaming service's broker system to distribute events
+        # Determine subgraphs parameter: if explicitly set, use that value; otherwise use False as default
+        subgraphs_param = False if subgraphs is None else subgraphs
```
ibbybuilds: @tokezooo you can just keep default value as false in the parameters and this won't be needed

src/agent_server/api/runs.py
```diff
@@ -650,6 +671,7 @@ async def execute_run_async(
     interrupt_before: Optional[Union[str, List[str]]] = None,
     interrupt_after: Optional[Union[str, List[str]]] = None,
     multitask_strategy: Optional[str] = None,
+    subgraphs: Optional[bool] = None,
```
ibbybuilds: make the default value `false`

src/agent_server/models/runs.py
```diff
@@ -43,6 +43,12 @@ class RunCreate(BaseModel):
         description="Nodes to interrupt immediately after they get executed. Use '*' for all nodes.",
     )
     
+    # Subgraph configuration
+    stream_subgraphs: Optional[bool] = Field(
+        None,
```
ibbybuilds: make the default value `false`

ibbybuilds (approved):

PR #34, opened 2025-09-13
Metadata for assistant, new count endpoint and assistant delete bug fix

### LINKED ISSUE #29: [FEATURE] Add metadata field to POST /assistants/search
**Is your feature request related to a problem? Please describe.**  
The `POST /assistants/search` endpoint is missing the `metadata` field, which breaks compatibility with the LangGraph platform.

**Describe the solution you'd like**  
Add support for the `metadata` field in the request and response of `POST /assistants/search`.

**Describe alternatives you've considered**  
No alternatives are available, since the field is required for compatibility with `langgraph dev`.

**Use Case**  
- Developers using `langgraph dev` to search for assistants.  
- Workflows that rely on assistant metadata to filter or manage assistants.  
- Ensures Aegra matches langgraph-api behavior.  

**Proposed Implementation**  
- Update request body to accept `metadata`.  
- Return `metadata` in the response payload.  
- Add relevant unit/integration tests.  

**Priority**  
- [ ] Low - Nice to have  
- [ ] Medium - Would significantly improve workflow  
- [x] High - Blocking current work  
- [ ] Critical - Required for production use  

**Additional context**  
Reference: [LangGraph API reference](https://langchain-ai.github.io/langgraph/cloud/reference/api/api_ref.html?#tag/assistants/post/assistants/search)

### LINKED ISSUE #30: [FEATURE] Add metadata field to POST /assistants (Create Assistant)
**Is your feature request related to a problem? Please describe.**  
The `POST /assistants` (Create Assistant) endpoint currently does not support the `metadata` field, which is required for full compatibility with langgraph-api.

**Describe the solution you'd like**  
Add support for the `metadata` field when creating assistants, and include it in the response object.

**Describe alternatives you've considered**  
No alternatives exist, since the field is necessary to replicate langgraph-api behavior and ensure assistants are created with complete attributes.

**Use Case**  
- Developers creating assistants via `langgraph dev`.  
- Applications requiring assistant metadata for filtering, categorization, or management.  
- Provides parity with langgraph-api.  

**Proposed Implementation**  
- Update `POST /assistants` to accept and persist the `metadata` field.  
- Return `metadata` in the response object.  
- Add tests for request/response validation.  

**Priority**  
- [ ] Low - Nice to have  
- [ ] Medium - Would significantly improve workflow  
- [x] High - Blocking current work  
- [ ] Critical - Required for production use  

**Additional context**  
Reference: [LangGraph API reference](https://langchain-ai.github.io/langgraph/cloud/reference/api/api_ref.html?#tag/assistants/post/assistants)

### LINKED ISSUE #33: [BUG] Delete assistant fails with FK violation when related runs exist
Calling DELETE `/assistants/{assistant_id}` throws a Postgres foreign key violation because runs.assistant_id references the assistant row being deleted and the FK is using the default NO ACTION/RESTRICT behavior. As a result, the delete is blocked when the assistant still has associated runs.

## Affected Code
```python
# src/agent_server/api/assistants.py
@router.delete("/assistants/{assistant_id}")
async def delete_assistant(
    assistant_id: str,
    user: User = Depends(get_current_user),
    session: AsyncSession = Depends(get_session),
):
    """Delete assistant by ID"""
    stmt = select(AssistantORM).where(
        AssistantORM.assistant_id == assistant_id, AssistantORM.user_id == user.identity
    )
    assistant = await session.scalar(stmt)

    if not assistant:
        raise HTTPException(404, f"Assistant '{assistant_id}' not found")

    await session.delete(assistant)
    await session.commit()
    return {"status": "deleted"}
```

## Error
```
(sqlalchemy.dialects.postgresql.asyncpg.IntegrityError)
<class 'asyncpg.exceptions.ForeignKeyViolationError'>:
update or delete on table "assistant" violates foreign key constraint
"runs_assistant_id_fkey" on table "runs"
DETAIL:  Key (assistant_id)=(company_research_1) is still referenced from table "runs".
[SQL: DELETE FROM assistant WHERE assistant.assistant_id = $1::VARCHAR]
[parameters: ('company_research_1',)]
(Background on this error at: https://sqlalche.me/e/20/gkpj)
```
SQLAlchemy is reflecting the DB rule correctly; Postgres forbids deleting a parent row while child rows reference it unless the FK specifies a delete action like `CASCADE` or `SET NULL`.

Steps to Reproduce

1. Create an assistant `A`.
2. Create one or more runs rows that reference `A.assistant_id`.
3. Call DELETE `/assistants/{assistant_id}` for A.
4. Observe ForeignKeyViolationError.

Expected: Assistant is deleted per our API contract (or rejected with a clear 409 if policy is to keep history).
Actual: 500 with FK violation; DB refuses deletion.

## Root Cause

`runs.assistant_id → assistant.assistant_id` foreign key is defined without an ON DELETE action (defaults to NO ACTION/RESTRICT). Deleting an assistant with existing runs therefore fails.

## Possible Fixes
Option A — Hard delete assistant and its runs (DB-level cascade)
Option B — Keep runs; detach them on delete
Option C — Soft delete assistants (preferable)

arthiondaena:
Closes #29, Closes #30 Added metadata for assistant and for assistant search. New count endpoint for counting the assistants. 
Fixes #33 Cascade delete for assistant_id foreign key in AssistantVersions and Runs table.

Updated langgraph-sdk because only versions > 0.2.4 supports count endpoint.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,6 @@ dependencies = [
     "sqlalchemy>=2.0.0",
     "uvicorn>=0.35.0",
     "langfuse>=3.3.4",
-    "langgraph-sdk==0.2.4",
 ]
 
 [project.urls]
@@ -88,6 +87,7 @@ markers = [
 dev = [
     "black>=25.1.0",
     "isort>=6.0.1",
+    "langgraph-sdk==0.2.4",
     "mypy>=1.17.1",
     "pytest>=8.4.1",
     "pytest-asyncio>=1.1.0",
--- a/src/agent_server/api/assistants.py
+++ b/src/agent_server/api/assistants.py
@@ -392,7 +392,8 @@ async def list_assistant_versions(
             user_id=user.identity,
             version=v.version,
             created_at=v.created_at,
-            updated_at=v.created_at
+            updated_at=v.created_at,
+            metadata_dict=v.metadata_dict
         ) for v in versions
     ]
 
--- a/uv.lock
+++ b/uv.lock
@@ -20,7 +20,6 @@ dependencies = [
     { name = "langfuse" },
     { name = "langgraph" },
     { name = "langgraph-checkpoint-postgres" },
-    { name = "langgraph-sdk" },
     { name = "psycopg", extra = ["binary"] },
     { name = "pydantic" },
     { name = "pyjwt" },
@@ -33,6 +32,7 @@ dependencies = [
 dev = [
     { name = "black" },
     { name = "isort" },
+    { name = "langgraph-sdk" },
     { name = "mypy" },
     { name = "pytest" },
     { name = "pytest-asyncio" },
@@ -50,7 +50,6 @@ requires-dist = [
     { name = "langfuse", specifier = ">=3.3.4" },
     { name = "langgraph", specifier = ">=0.6.0" },
     { name = "langgraph-checkpoint-postgres", specifier = ">=2.0.23" },
-    { name = "langgraph-sdk", specifier = "==0.2.4" },
     { name = "psycopg", extras = ["binary"], specifier = ">=3.2.9" },
     { name = "pydantic", specifier = ">=2.11.7" },
     { name = "pyjwt", specifier = ">=2.10.1" },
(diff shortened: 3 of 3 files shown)


pyproject.toml
```diff
@@ -41,6 +41,7 @@ dependencies = [
     "sqlalchemy>=2.0.0",
     "uvicorn>=0.35.0",
     "langfuse>=3.3.4",
+    "langgraph-sdk==0.2.4",
```
ibbybuilds: we use `langgraph-sdk` only for e2e tests as a client, wouldn't this be better placed in dev dependencies or test to keep the env lean?

arthiondaena: Got it, pushed the changes.

ibbybuilds (approved):
ok

Issue #39, opened 2025-09-25
[BUG] Langgraph studio integration


JuanjoInsud:
**Describe the bug**
[Langgraph studio](https://github.com/langchain-ai/langgraph-studio) doen't work with the aegra api. 
https://smith.langchain.com/studio/thread?baseUrl=http%3A%2F%2F127.0.0.1%3A8000

It gives an Entity 422 Error 
when the client made the post assistants/search with payload:
"{\"limit\":10,\"offset\":0,\"metadata\":{\"created_by\":\"system\"}}"

**To Reproduce**
1. docker-compose-up
2. Open https://smith.langchain.com/studio/thread?baseUrl=http%3A%2F%2F127.0.0.1%3A2024 in web
3. Look for network requests in dev tools

**Expected behavior**
Load the graphs and assisants. Configure them also. Run any assistant throw the api.

**Actual behavior**
It gives an Entity 422 Error 
when the client made the post assistants/search with payload:
"{\"limit\":10,\"offset\":0,\"metadata\":{\"created_by\":\"system\"}}"

**Possible Solution**
Fix the api endpoints definitions

fduches:
+1
fix please

arthiondaena:
Features required to support studio integration

- [ ] Support text type request body for routes.
- [ ] Implement `/graphs` and `/subgraphs` endpoints.
- [ ] Implement `threads/{thread_id}/state/{checkpoint_id}` and `threads/{thread_id}/state/checkpoint routes`.
- [ ] Change server version in `/info` endpoint from `0.1.0` to `0.4.35`.
- [ ] Create initial version for default graphs.
- [ ] Edit `/versions` endpoint to add `"system"` user_id.

ibbybuilds:
@fduches @JuanjoInsud @tokezooo LangGraph studio is now supported please check and report if you find any bugs.
Resolved in #102

PR #65, opened 2025-10-13
feat: support assistant settings and implement structlog

### LINKED ISSUE #19: [FEATURE] Use the Assistant Context if it is not provided during runtime
Problem:
When creating an Assistant, we can provide the config/context to the graph. But when we try to use an assistant in like Agent Chat App, we cannot provide `config/context` for very /run/stream API call. So, if there is a config/context attached to Assistant during it's creation, it should be leveraged. 

Solution:
A simple `if` block to assign the assistants context during a `run`.

- API changes needed

- [ ] Low - Nice to have

arthiondaena:
## Type of Change

- [x] `feat`: New feature
- [ ] `fix`: Bug fix
- [x] `refactor`: Code refactoring
- [ ] `ci`: CI/CD changes

## Related Issues

Related #39 
Closes #19 

## Changes Made

- Fixed all the issues related to agent settings in langsmith, which includes
  - Versioning for default assistants
  - Flags required by langsmith to access assistant settings. (`/info` endpoint)
  - Use assistant config in thread as default config if config/context is not explicitly sent for run creation.
- Replace logging module with structlog, which is more suitable for production environment.x

## Testing

- [ ] E2E tests added/updated

## Checklist

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,17 @@
+coverage:
+  status:
+    # Fail PRs that reduce total coverage by more than 2%
+    project:
+      default:
+        target: auto
+        threshold: 2%
+    # Treat patch coverage as informational only
+    patch:
+      default:
+        informational: true
+
+comment:
+  hide_project_coverage: false
+
+# When modifying this file, please validate using
+# curl -X POST --data-binary @codecov.yml https://codecov.io/validate
--- a/src/agent_server/observability/base.py
+++ b/src/agent_server/observability/base.py
@@ -1,0 +1,100 @@
+"""Base observability interface for extensible tracing and monitoring."""
+
+import logging
+from abc import ABC, abstractmethod
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+
+class ObservabilityProvider(ABC):
+    """Abstract base class for observability providers."""
+
+    @abstractmethod
+    def get_callbacks(self) -> list[Any]:
+        """Return a list of callbacks for this provider."""
+        pass
+
+    @abstractmethod
+    def get_metadata(
+        self, run_id: str, thread_id: str, user_identity: str | None = None
+    ) -> dict[str, Any]:
+        """Return metadata to be added to the run configuration."""
+        pass
+
+    @abstractmethod
+    def is_enabled(self) -> bool:
+        """Check if this provider is enabled."""
+        pass
+
+
+class ObservabilityManager:
+    """Manages multiple observability providers."""
+
+    def __init__(self) -> None:
+        self._providers: list[ObservabilityProvider] = []
+
+    def register_provider(self, provider: ObservabilityProvider) -> None:
+        """Register an observability provider (idempotent for same instance).
+
+        Only registers enabled providers. Skips if the exact same instance
+        is already registered (checked by object identity).
+        """
+        if not provider.is_enabled():
+            return
+
+        # Check if this exact instance is already registered (by object identity)
+        if provider in self._providers:
+            return
+
+        self._providers.append(provider)
+
+    def get_all_callbacks(self) -> list[Any]:
+        """Get callbacks from all enabled providers."""
+        callbacks = []
+        for provider in self._providers:
+            try:
+                callbacks.extend(provider.get_callbacks())
+            except Exception as e:
+                logger.error(
+                    f"Failed to get callbacks from {provider.__class__.__name__}: {e}"
+                )
+        return callbacks
+
+    def get_all_metadata(
+        self, run_id: str, thread_id: str, user_identity: str | None = None
+    ) -> dict[str, Any]:
+        """Get metadata from all enabled providers."""
+        metadata = {}
+        for provider in self._providers:
+            try:
+                provider_metadata = provider.get_metadata(
+                    run_id, thread_id, user_identity
+                )
+                metadata.update(provider_metadata)
+            except Exception as e:
+                logger.error(
+                    f"Failed to get metadata from {provider.__class__.__name__}: {e}"
+                )
+        return metadata
+
+
+# Global observability manager instance
+_observability_manager = ObservabilityManager()
+
+
+def get_observability_manager() -> ObservabilityManager:
+    """Get the global observability manager."""
+    return _observability_manager
+
+
+def get_tracing_callbacks() -> list[Any]:
+    """Get callbacks from all registered observability providers."""
+    return _observability_manager.get_all_callbacks()
+
+
+def get_tracing_metadata(
+    run_id: str, thread_id: str, user_identity: str | None = None
+) -> dict[str, Any]:
+    """Get metadata from all registered observability providers."""
+    return _observability_manager.get_all_metadata(run_id, thread_id, user_identity)
--- a/src/agent_server/observability/langfuse_integration.py
+++ b/src/agent_server/observability/langfuse_integration.py
@@ -1,31 +1,81 @@
 import os
+from typing import Any
 
 import structlog
+
+from .base import ObservabilityProvider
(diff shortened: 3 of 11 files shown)


src/agent_server/api/runs.py
```diff
@@ -145,6 +154,19 @@ async def create_run(
...
+        raise HTTPException(
+            status_code=400,
+            detail="Cannot specify both configurable and context. Prefer setting context alone. Context was introduced in LangGraph 0.6.0 and is the long term planned replacement for configurable.",
+        )
+
+    if context:
+        configurable = context.copy()
+        config["configurable"] = configurable
+    else:
+        context = configurable.copy()
```
ibbybuilds, line 160: what if context schema is different from configurable? because the reason for removing it was that on langgraph sutdio the schema for context was not matching (there was thread_id coming in config which was not present in context schema). have you tested this with langgraph studio?

src/agent_server/api/runs.py
```diff
@@ -104,6 +104,15 @@ async def update_thread_metadata(
     await session.commit()
 
 
+def _merge_jsonb(*objects: dict) -> dict:
```
ibbybuilds: better to move the method as utility in utils file

src/agent_server/core/health.py
```diff
@@ -35,6 +36,12 @@ async def info() -> InfoResponse:
         version="0.1.0",
         description="Production-ready Agent Protocol server built on LangGraph",
         status="running",
+        flags={
+            "assistants": True,
+            "crons": False,
+            "langsmith": True,
```
ibbybuilds: why is langsmith true here, we don't support it yet i think?

src/agent_server/core/health.py
```diff
@@ -35,6 +36,12 @@ async def info() -> InfoResponse:
...
+            "langsmith_tracing_replicas": True,
```
ibbybuilds: same here

run_server.py
```diff
@@ -68,17 +75,18 @@ def main():
...
-    print("🧪 Test with: python test_sdk_integration.py")
+    logger.info("🚀 Starting Aegra...")
+    logger.info(f"📍 Server will be available at: http://localhost:{port}")
+    logger.info(f"📊 API docs will be available at: http://localhost:{port}/docs")
+    logger.info("🧪 Test with: python test_sdk_integration.py")
 
     uvicorn.run(
         "src.agent_server.main:app",
-        host="0.0.0.0",
+        host=os.getenv("HOST", "127.0.0.1"),
```
ibbybuilds: **Problem:**
Changing the default host from `0.0.0.0` to `127.0.0.1` breaks Docker containers and production deployments. Containers need `0.0.0.0` to accept external connections.

**Current code:**
```python
host=os.getenv("HOST", "127.0.0.1"),
```

**Solution:**
```python
host=os.getenv("HOST", "0.0.0.0"),  # Restore original default
```

This maintains backward compatibility while still allowing override via env var.

src/agent_server/main.py
```diff
@@ -20,8 +20,8 @@
     sys.path.insert(0, str(graphs_dir))
 
 # ruff: noqa: E402 - imports below require sys.path modification above
```
ibbybuilds, line 22: **Problem:**
`setup_logging()` is only called in `run_server.py`, but not in `main.py`. When running the app directly via uvicorn (e.g., `uvicorn src.agent_server.main:app`), structlog is never configured, causing logging to fail or use wrong format.

**Current code:**
```python
import structlog
logger = structlog.getLogger(__name__)
```

**Solution:**
```python
import structlog
from .utils.setup_logging import setup_logging

# Configure logging at module level
setup_logging()
logger = structlog.getLogger(__name__)
```

This ensures logging works regardless of entry point.

src/agent_server/utils/setup_logging.py
```diff
@@ -0,0 +1,129 @@
+import logging
+import logging.config
+import os
+from typing import Any
+
+import structlog
+
+
+class InterceptHandler(logging.Handler):
```
ibbybuilds: ### ⚠️ Dead code: InterceptHandler is never used

**Problem:**
This handler is defined but never used anywhere. Additionally, the implementation just calls `logger.handle(record)` on the same record, which could cause infinite recursion.

**Current code:**
```python
class InterceptHandler(logging.Handler):
    def emit(self, record: logging.LogRecord) -> None:
        logger = logging.getLogger(record.name)
        logger.handle(record)
```
**Solution:**
Either remove it entirely, or implement it properly to bridge stdlib → structlog:

```python
class InterceptHandler(logging.Handler):
    """Intercept standard logging and forward to structlog"""
    def emit(self, record: logging.LogRecord) -> None:
        logger = structlog.get_logger(record.name)
        level = record.levelname.lower()
        getattr(logger, level)(record.getMessage(), exc_info=record.exc_info)
```

src/agent_server/api/runs.py
arthiondaena: This is exactly how the original langgraph is implemented.
And yes, langgraph studio perfectly works with this implementation. You can try to test it once, and let me know if you encounter any error.
In our case context schema will differ from configurable, since we will be injecting thread_id and other values. Context schema must only have key/value pairs which are defined in Context class ([`graphs/react_agent/context.py`](https://github.com/ibbybuilds/aegra/blob/1bbaf808bcb628259d4d8756e5d6c3537fca3d36/graphs/react_agent/context.py#L13))

src/agent_server/api/runs.py
arthiondaena: Got it

run_server.py
arthiondaena: The reason for this was because of pre-commit hook. It's getting failed because of hardcoded bind to all interfaces.
```
check for merge conflicts................................................Passed
bandit...................................................................Failed
- hook id: bandit
- exit code: 1

[main]  INFO    profile include tests: None
[main]  INFO    profile exclude tests: B101
[main]  INFO    cli include tests: None
[main]  INFO    cli exclude tests: None
[main]  INFO    using config: pyproject.toml
[main]  INFO    running on Python 3.11.13
Run started:2025-10-14 14:58:54.868663

Test results:
>> Issue: [B104:hardcoded_bind_all_interfaces] Possible binding to all interfaces.
   Severity: Medium   Confidence: Medium
   CWE: CWE-605 (https://cwe.mitre.org/data/definitions/605.html)
   More Info: https://bandit.readthedocs.io/en/0.0.0/plugins/b104_hardcoded_bind_all_interfaces.html
   Location: .\run_server.py:85:31
84              "src.agent_server.main:app",
85              host=os.getenv("HOST", "0.0.0.0"),
86              port=port,

--------------------------------------------------

Code scanned:
        Total lines of code: 66
        Total lines skipped (#nosec): 0
        Total potential issues skipped due to specifically being disabled (e.g., #nosec BXXX): 0

Run metrics:
        Total issues (by severity):
                Undefined: 0
                Low: 0
                Medium: 1
                High: 0
        Total issues (by confidence):
                Undefined: 0
                Low: 0
                Medium: 1
                High: 0
Files skipped (0):
```

arthiondaena: Should I disable ruff for this line?

ibbybuilds: `# nosec B104 - required for Docker`

yes add this

src/agent_server/utils/run_utils.py
```diff
@@ -0,0 +1,31 @@
+import copy
+from typing import Any
+
+
+def _should_skip_event(raw_event: Any) -> bool:
```
ibbybuilds, line 5: these are not private methods anymore, please remove the underscrore prefix

PR #83, opened 2025-11-04
feat: implement `threads/:thread_id/state` endpoint to get a thread's state

pharesdiego:
## Description
Implements `threads/:thread_id/state` to make it possible to get a thread's state (i.e. latest checkpoint). This endpoint is defined in [langgraph's api specs](https://langchain-ai.github.io/langgraph/cloud/reference/api/api_ref.html#tag/threads/get/threads/{thread_id}/state) and accessible through the sdk calling `client.threads.get_state`.

The `threads/:thread_id/state` endpoint is the one used by the sdk when `checkpoint` or `checkpoint_id` are not defined as arguments to `client.threads.get_state`.

## Type of Change

- [x] `feat`: New feature
- [ ] `fix`: Bug fix
- [ ] `ci`: CI/CD changes

## Changes Made

- Added new `threads/:thread_id/state` endpoint.
- Added new `subgraph_hitl_agent` to test agents with subgraphs and HITL approaches.
- Subgraphs state are now included in the `tasks` array when `subgraphs` argument is set to `True` instead of setting it to `None`.

## Testing

- [x] E2E tests added/updated

## Checklist

## Additional Notes

- LangGraph only exposes subgraph checkpoints while the run is [interrupted](https://docs.langchain.com/oss/python/langgraph/use-subgraphs#view-subgraph-state).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -19,7 +19,9 @@ from tests.fixtures.langgraph import (
     FakeAgent,
     FakeGraph,
     FakeSnapshot,
+    make_interrupt,
     make_snapshot,
+    make_task,
     patch_langgraph_service,
 )
 from tests.fixtures.session_fixtures import (
@@ -45,7 +47,9 @@ __all__ = [
     "FakeSnapshot",
     "FakeAgent",
     "FakeGraph",
+    "make_interrupt",
     "make_snapshot",
+    "make_task",
     "patch_langgraph_service",
     "create_test_app",
     "make_client",
--- a/tests/fixtures/langgraph.py
+++ b/tests/fixtures/langgraph.py
@@ -3,6 +3,8 @@
 from typing import Any
 from unittest.mock import patch
 
+from langgraph.types import Interrupt, PregelTask
+
 
 class FakeSnapshot:
     """Mock LangGraph snapshot"""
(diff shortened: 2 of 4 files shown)


src/agent_server/services/thread_state_service.py
```diff
@@ -30,6 +30,21 @@ def convert_snapshot_to_thread_state(
...
+                    if "state" in task and task["state"] is not None:
+                        try:
+                            task["state"] = self.convert_snapshot_to_thread_state(
+                                task["state"], thread_id, subgraphs=True
+                            )
+                        except Exception as e:
+                            logger.error(
+                                f"Failed to serialize subgraph state for task {task.get('id')}: {e}"
+                            )
+                            task["state"] = None
```
pharesdiego, line 46: This should be handled in [`LanggraphSerializer.serialize_task`](https://github.com/pharesdiego/aegra/blob/3a15ec951d80def7a6aa859fef486bc935f6a201/src/agent_server/core/serializers/langgraph.py#L34) instead but using `convert_snapshot_to_thread_state` there would create a circular dependency between `ThreadStateService` and `LanggraphSerializer`.

A solution would be to create a new serializer class that holds `convert_snapshot_to_thread_state` so that it can be shared between the other two serializers.

PR #149, opened 2026-01-29
fix(services): prevent dual-yield in get_graph context manager

### LINKED ISSUE #148: [BUG] "generator didn't stop after athrow()" in get_graph context manager
## Description

The `get_graph()` async context manager in `langgraph_service.py` has a dual-yield pattern that violates Python's async generator protocol:

```python
@asynccontextmanager
async def get_graph(self, graph_id: str) -> AsyncIterator[Pregel]:
    ...
    try:
        yield base_graph.copy(update={"checkpointer": checkpointer, "store": store})
    except Exception:
        logger.warning(...)
        yield base_graph  # <-- BUG: second yield after athrow()
```

When an exception occurs in the context body (e.g., Pydantic validation fails during graph execution), `@asynccontextmanager` calls `athrow()` on the generator. The `except Exception` catches this, then tries to yield again, causing:

```
RuntimeError: generator didn't stop after athrow()
```

## Steps to Reproduce

1. Create a graph with Pydantic state that uses structured output
2. Have the LLM return invalid data that fails Pydantic validation
3. The exception during parsing triggers the bug

## Expected Behavior

Exception should propagate normally without a secondary "generator didn't stop" error.

## Suggested Fix

Move the try/except BEFORE the yield:

```python
# Try to create a copy with checkpointer/store injected
try:
    graph_to_use = base_graph.copy(update={"checkpointer": checkpointer, "store": store})
except Exception:
    logger.warning(
        f"⚠️  Graph '{graph_id}' does not support checkpointer injection; "
        "running without persistence"
    )
    graph_to_use = base_graph

yield graph_to_use  # Single yield
```

## Environment

- aegra version: main branch (commit 3ec8d68)
- Python: 3.13
- LangGraph: latest

Happy to submit a PR with this fix if you'd like.

SlanyCukr:
## Summary

Move the try/except for `copy()` BEFORE the yield statement to prevent "generator didn't stop after athrow()" errors.

## Problem

When an exception occurs in the context body (e.g., Pydantic validation fails during graph execution), `@asynccontextmanager` calls `athrow()` on the generator. With the previous pattern:

```python
try:
    yield base_graph.copy(...)
except Exception:
    yield base_graph  # <-- BUG: second yield after athrow()
```

The `except` clause catches the exception thrown via `athrow()` and tries to yield again, violating async generator protocol.

## Solution

Do the copy BEFORE the yield, so there's only ever one yield:

```python
try:
    graph_to_use = base_graph.copy(...)
except Exception:
    graph_to_use = base_graph

yield graph_to_use  # Single yield
```

## Test Plan

- Existing tests should pass (behavior unchanged for happy path)
- Graph execution errors now propagate cleanly without secondary "generator didn't stop" error

Fixes #148

### DIFF
--- a/src/agent_server/services/langgraph_service.py
+++ b/src/agent_server/services/langgraph_service.py
@@ -251,16 +251,23 @@ class LangGraphService:
         checkpointer = db_manager.get_checkpointer()
         store = db_manager.get_store()
 
-        # Yield a fresh copy with checkpointer/store injected
+        # Try to create a copy with checkpointer/store injected.
+        # NOTE: Do this BEFORE yield to avoid dual-yield when exceptions occur
+        # in the context body - @asynccontextmanager would call athrow() and
+        # catch it in except, causing "generator didn't stop after athrow()".
         try:
-            yield base_graph.copy(update={"checkpointer": checkpointer, "store": store})
+            graph_to_use = base_graph.copy(
+                update={"checkpointer": checkpointer, "store": store}
+            )
         except Exception:
-            # Fallback: property may be immutably set
+            # Graph doesn't support copy with these attrs (e.g., immutable property)
             logger.warning(
                 f"⚠️  Graph '{graph_id}' does not support checkpointer injection; "
                 "running without persistence"
             )
-            yield base_graph
+            graph_to_use = base_graph
+
+        yield graph_to_use
 
     async def get_graph_for_validation(self, graph_id: str) -> Pregel:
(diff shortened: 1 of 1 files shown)


Witaly3 (commented):
Looks good. Separating the `copy()` logic from the `yield` is the right way to handle exceptions here without breaking the async generator protocol.

ibbybuilds (approved):
thanks for catching this

PR #216, opened 2026-02-28
fix(api): correctly store and replay updates, debug, custom, and messages events

Witaly3:
## Description

  This PR fixes four related bugs in the SSE streaming pipeline that affect event replay (reconnect) and certain stream modes.

  The streaming pipeline has two phases:
  1. **Live streaming** — raw events from LangGraph are pushed to an in-memory broker and simultaneously stored in the event store (DB).
  2. **Replay** — when a client reconnects with `Last-Event-ID`, stored events are fetched from DB and re-serialized back to SSE format.

  All four bugs manifest in phase 2 (replay) or in incorrect SSE output for certain stream modes.

  ---

  ### Bug 1: `stream_mode=updates` events were stored under the `values` event type

  **Root cause:** In `streaming_service.py`, `store_event_from_raw` had a single `elif` branch for both `values` and `updates`:

  ```python
  elif stream_mode_label == "values" or stream_mode_label == "updates":
      await store_sse_event(run_id, event_id, "values", ...)  # ← hardcoded "values"
  ```

  Both modes were written to the event store with event type `"values"`. On reconnect, a client that requested `stream_mode=["updates"]` would receive `event:
  values` SSE events instead of `event: updates`. This breaks clients that differentiate between update deltas and full state snapshots — e.g. any UI relying on the
  LangGraph SDK's typed stream would receive the wrong event type.

  **Fix:** Split into two separate branches — `values` stores `"values"` and `updates` stores `"updates"`.

  ---

  ### Bug 2: `debug`, `custom`, and `metadata` events were silently dropped and never stored

  **Root cause:** `store_event_from_raw` had no `elif` branch for `debug`, `custom`, or `metadata` stream modes. These events were emitted to the live broker for
  current subscribers, but never persisted. Any client that reconnected mid-stream or retrieved the run history after completion would receive an incomplete stream —
   all `debug`, `custom`, and `metadata` events would be missing.

  **Fix:** Added dedicated storage branches for each mode:
  - `debug` → stored as `{"debug": payload}` (consistent with how `convert_stored_to_sse` reads it via `create_debug_event`)
  - `custom` → stored as `{"chunk": payload}` (consistent with the generic replay handler)
  - `metadata` → stored as the payload dict itself (the run-level metadata object emitted by LangGraph at stream start)

  ---

  ### Bug 3: `messages/partial`, `messages/complete`, `messages/metadata` produced malformed SSE on replay

  **Root cause:** `EventConverter.convert_stored_to_sse` had no explicit handler for these three event types. They fell through to the generic handler:

  ```python
  payload = data.get(event_type) or data.get("chunk") or data
  ```

  For `messages/partial`, `data.get("messages/partial")` returns `None` (that key doesn't exist in the stored dict), `data.get("chunk")` also `None`, so `data` — the
   entire stored envelope `{"type": "messages_partial", "messages": [...], "node_path": ...}` — was sent as the SSE payload. A replayed `messages/partial` event
  would contain an extra `type` and `node_path` wrapper that the live stream never had, breaking client-side message accumulation.

  **Fix:** Added explicit branches that extract the inner payload by key before serializing:
  - `messages/partial` and `messages/complete` → extract `data["messages"]`
  - `messages/metadata` → extract `data["metadata"]`

  This matches exactly what LangGraph emits in the live stream.

  ---

  ### Bug 4: `EventConverter` was double-remapping `__interrupt__` payloads in `updates` mode

  **Root cause:** `EventConverter._create_sse_event` contained logic to detect `__interrupt__` inside `updates` payloads and silently re-label them as `values`
  events:

  ```python
  if stream_mode == "updates":
      if isinstance(payload, dict) and "__interrupt__" in payload:
          event_type = "values"  # ← forced remapping
          return format_sse_message(event_type, payload, event_id)
  ```

  This remapping already happens upstream in `graph_streaming` — only for the case where `updates` was **not** explicitly requested by the client. When a user
  explicitly subscribes to `stream_mode=["updates"]`, interrupt events are LangGraph graph state updates and should be delivered as `updates`. The converter was
  blindly converting them to `values` regardless, meaning:

  - Clients that explicitly requested `stream_mode=["updates"]` received their interrupt events mislabeled as `values`.
  - With subgraphs enabled, the namespace prefix was also wrong (`values|subagent` instead of `updates|subagent`).

  **Fix:** Removed the interrupt detection from `EventConverter`. The converter now passes all `updates` payloads through as-is. The comment explains that interrupt
  remapping is the responsibility of `graph_streaming`, which already handles it before events reach the converter.

  ---

  ## Type of Change

  - [ ] `feat`: New feature
  - [x] `fix`: Bug fix
  - [ ] `ci`: CI/CD changes

  ## Related Issues

  

  ## Changes Made

  - **`streaming_service.py`**: Split the merged `values`/`updates` storage branch into two separate branches; added storage branches for `debug`, `custom`, and
  `metadata` stream modes that were previously unhandled
  - **`event_converter.py`**: Added explicit `convert_stored_to_sse` handlers for `messages/partial`, `messages/complete`, and `messages/metadata` that extract the
  inner payload correctly; removed the incorrect `__interrupt__` → `values` remapping from `_create_sse_event` (this responsibility belongs to `graph_streaming`, not
   the converter)

  ## Testing

  - [ ] E2E tests added/updated

  **Unit tests added:**

  - `test_store_event_from_raw_updates` — verifies `updates` events are stored with `"updates"` event type
  - `test_store_event_from_raw_updates_not_stored_as_values` — regression guard: asserts the stored event type is never `"values"` for an `updates` input
  - `test_store_event_from_raw_debug` — verifies `debug` events are stored and wrapped correctly
  - `test_store_event_from_raw_custom` — verifies `custom` events are stored with `{"chunk": ...}` envelope
  - `test_store_event_from_raw_run_metadata` — verifies `metadata` events are stored as-is
  - `test_convert_stored_to_sse_messages_partial` — verifies replay of `messages/partial` emits only the messages list, not the storage envelope
  - `test_convert_stored_to_sse_messages_complete` — same for `messages/complete`
  - `test_convert_stored_to_sse_messages_metadata` — same for `messages/metadata`, verifies inner metadata dict is in the SSE output
  - `test_convert_stored_to_sse_updates` — verifies replayed `updates` event produces `event: updates`, not `event: values`
  - `test_create_sse_event_updates_with_interrupt` — updated: interrupt payloads in `updates` mode must remain `event: updates`
  - `test_create_sse_event_updates_with_interrupt_and_namespace` — updated: with subgraphs enabled, interrupt in `updates` mode must be `event: updates|subagent`,
  not `values|subagent`

  ## Checklist

  ## Additional Notes

  All four bugs are only observable in the **replay path** (client reconnect or fetching a completed run's stream) or when using specific stream modes (`updates`,
  `debug`, `custom`, `metadata`, `messages/partial`, `messages/complete`, `messages/metadata`). The live streaming path for the common `values` + `messages` case was
   unaffected. The fixes are purely additive storage branches and targeted converter routing — no existing behavior for already-working paths is changed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Streamed events are categorized and preserved more explicitly: messages (partial/complete/metadata), updates (preserved as updates), debug, custom, values, end, and errors. Metadata replay uses the stored payload (including run/attempt); updates and custom chunks are passed through unchanged.

* **Tests**
  * Expanded unit coverage for messages, updates, debug, custom, metadata, namespaces, replay semantics, and edge cases (including regression checks).

### DIFF
--- a/libs/aegra-api/src/aegra_api/services/event_converter.py
+++ b/libs/aegra-api/src/aegra_api/services/event_converter.py
@@ -28,7 +28,7 @@ class EventConverter:
         stream_mode, payload, namespace = self._parse_raw_event(raw_event)
         return self._create_sse_event(stream_mode, payload, event_id, namespace)
 
-    def convert_stored_to_sse(self, stored_event: SSEEvent, run_id: str | None = None) -> str | None:
+    def convert_stored_to_sse(self, stored_event: SSEEvent, _run_id: str | None = None) -> str | None:
         """Convert stored event to SSE format"""
         event_type = stored_event.event
         data = stored_event.data
@@ -47,7 +47,10 @@ class EventConverter:
             # to what LangGraph originally emitted (run_id, attempt, etc.)
             return format_sse_message(event_type, data, event_id)
         elif event_type == "debug":
-            return create_debug_event(data.get("debug"), event_id)
+            debug_payload = data.get("debug")
+            if debug_payload is None:
+                return
+            return create_debug_event(debug_payload, event_id)
         elif event_type in ("messages/partial", "messages/complete"):
             messages = data.get("messages")
             if messages is None:
--- a/libs/aegra-api/src/aegra_api/services/streaming_service.py
+++ b/libs/aegra-api/src/aegra_api/services/streaming_service.py
@@ -381,9 +381,9 @@ class StreamingService:
         """Clean up streaming resources for a run"""
         broker_manager.cleanup_broker(run_id)
 
-    def _stored_event_to_sse(self, run_id: str, ev) -> str | None:
+    def _stored_event_to_sse(self, _run_id: str, ev) -> str | None:
         """Convert stored event object to SSE string"""
-        return self.event_converter.convert_stored_to_sse(ev, run_id)
+        return self.event_converter.convert_stored_to_sse(ev)
 
 
 # Global streaming service instance
--- a/libs/aegra-api/tests/unit/test_services/test_event_converter.py
+++ b/libs/aegra-api/tests/unit/test_services/test_event_converter.py
@@ -237,7 +237,7 @@ class TestEventConverter:
         stored_event.data = {"run_id": "run-123", "attempt": 1}
         stored_event.id = "evt-1"
 
-        result = self.converter.convert_stored_to_sse(stored_event, run_id="run-123")
+        result = self.converter.convert_stored_to_sse(stored_event)
 
         assert result is not None
         assert "event: metadata\n" in result
@@ -481,7 +481,7 @@ class TestEventConverter:
         stored_event.data = {"run_id": "run-abc", "attempt": 3}
         stored_event.id = "evt-1"
 
-        result = self.converter.convert_stored_to_sse(stored_event, run_id="run-abc")
+        result = self.converter.convert_stored_to_sse(stored_event)
 
         assert result is not None
         assert "event: metadata\n" in result


libs/aegra-api/src/aegra_api/services/event_converter.py
```diff
@@ -46,6 +46,14 @@ def convert_stored_to_sse(self, stored_event, run_id: str = None) -> str | None:
             return create_metadata_event(run_id, event_id)
         elif event_type == "debug":
             return create_debug_event(data.get("debug"), event_id)
+        elif event_type in ("messages/partial", "messages/complete"):
+            # Extract the messages list stored under the "messages" key
+            messages = data.get("messages")
+            return format_sse_message(event_type, messages, event_id)
+        elif event_type == "messages/metadata":
```
ibbybuilds, line 59: **Guard against `None`:** If stored data is malformed and lacks a `"messages"` key, `data.get("messages")` returns `None`, which gets passed to `format_sse_message` as the payload. Consider adding a null guard consistent with the existing `messages` handler pattern (line 38-44 has `if message_chunk is None: return None`):

```python
messages = data.get("messages")
if messages is None:
    return None
return format_sse_message(event_type, messages, event_id)
```

Same applies to `data.get("metadata")` on line 56.

libs/aegra-api/src/aegra_api/services/event_converter.py
```diff
@@ -46,6 +46,14 @@ def convert_stored_to_sse(self, stored_event, run_id: str = None) -> str | None:
             return create_metadata_event(run_id, event_id)
```
ibbybuilds: **Pre-existing issue worth noting:** This `metadata` handler calls `create_metadata_event(run_id, event_id)` which reconstructs `{"run_id": run_id, "attempt": 1}` from scratch. The metadata payload you now carefully store in `streaming_service.py` (line 175) is **completely ignored** on replay. The storage and replay are asymmetric — the stored content is never used.

This isn't something your PR introduced (the handler existed before), but the new storage code creates a false sense that the stored metadata will be faithfully replayed. Worth a follow-up to either:
- Have the replay handler use the stored data, or
- Document why it intentionally reconstructs from scratch

libs/aegra-api/src/aegra_api/services/event_converter.py
```diff
@@ -46,6 +46,14 @@ def convert_stored_to_sse(self, stored_event, run_id: str = None) -> str | None:
...
+            # Extract the messages list stored under the "messages" key
+            messages = data.get("messages")
+            return format_sse_message(event_type, messages, event_id)
+        elif event_type == "messages/metadata":
+            # Extract the metadata stored under the "metadata" key
+            metadata = data.get("metadata")
+            return format_sse_message("messages/metadata", metadata, event_id)
         elif event_type == "end":
             return create_end_event(event_id)
         elif event_type == "error":
```
ibbybuilds, line 69: **`updates` and `custom` event replay rely on the generic fallback — consider explicit branches.**

When a stored `updates` event is replayed, there is no `elif event_type == "updates"` branch here. It falls through to the generic handler (line 62+):
```python
payload = data.get(event_type) or data.get("chunk") or data
```
`data.get("updates")` → `None`, then `data.get("chunk")` → the payload. Works by accident.

Same for `custom`: `data.get("custom")` → `None`, falls to `data.get("chunk")`.

This is fragile. Adding explicit branches would be consistent with the explicit handling you added for `messages/*` above:

```python
elif event_type == "updates":
    payload = data.get("chunk") or data
    return format_sse_message("updates", payload, event_id)
elif event_type == "custom":
    payload = data.get("chunk") or data
    return format_sse_message("custom", payload, event_id)
```

libs/aegra-api/src/aegra_api/services/streaming_service.py
```diff
@@ -135,13 +135,41 @@ async def store_event_from_raw(
...
+        elif stream_mode_label == "metadata":
+            await store_sse_event(
+                run_id,
+                event_id,
+                "metadata",
+                event_payload if isinstance(event_payload, dict) else {},
+            )
         elif stream_mode_label == "end":
             await store_sse_event(
                 run_id,
```
ibbybuilds, line 175: See the note on `event_converter.py` line 46 — this stored metadata payload is currently ignored on replay. The `convert_stored_to_sse` handler for `"metadata"` calls `create_metadata_event(run_id, event_id)` which reconstructs from scratch. Not a blocker but worth being aware of the asymmetry.

PR #222, opened 2026-03-03
feat(observability): propagate user, session, and graph metadata to OTEL traces

### LINKED ISSUE #213: How to set LangFuse user_id, session_id, and trace name via OTEL?
We're using Aegra with `OTEL_TARGETS="LANGFUSE"` and traces are flowing correctly. However, we'd like to enrich those traces with:

1. **`user_id`** -- we have the authenticated user identity available via `config["configurable"]["langgraph_auth_user"]["identity"]` inside our agents
2. **`session_id`** -- we'd like to map this to the LangGraph `thread_id`
3. **Agent/trace naming** -- all our graphs currently show up as "LangGraph" in LangFuse, even though we pass `name=` to `create_agent()`. We'd like each graph (`matter_agent`, `research_agent`, `drafting_agent`, `title_agent`) to appear with its own name.

**Our current plan** is to use the `opentelemetry-api` directly inside agent middleware to set span attributes that LangFuse's OTLP endpoint recognises:

```python
from opentelemetry import trace

span = trace.get_current_span()
if span.is_recording():
    span.set_attribute("langfuse.user.id", user_id)
    span.set_attribute("langfuse.session.id", thread_id)
```

This is based on [LangFuse's OTEL property mapping](https://langfuse.com/docs/opentelemetry#property-mapping).

**Questions:**

1. Is this the recommended way to attach custom metadata to OTEL traces in Aegra, or is there a built-in mechanism we're missing (e.g. something in `aegra.json`, an env var, or a hook)?
2. Does the `name` parameter on `create_agent()` / `StateGraph.compile()` flow through to the OTEL root span name, or do we need to set `langfuse.trace.name` explicitly?
3. Are there any plans for first-class support for propagating `user_id` / `session_id` from Aegra's auth system into OTEL traces automatically?

**Environment:**
- `aegra-cli>=0.3.3`
- `langgraph>=1.0.8`
- Python 3.11
- `OTEL_TARGETS="LANGFUSE"`

Thanks!

Witaly3:
## Description

  Fixes a long-standing gap in Aegra's OpenTelemetry integration: when using `OTEL_TARGETS=LANGFUSE`
  (or `PHOENIX`), traces contained no user identity, no session mapping, and all graphs appeared under
  the generic name "LangGraph" regardless of which graph was actually executed.

  This PR introduces `SpanEnrichmentProcessor` — a lightweight per-request `SpanProcessor` that
  automatically injects the correct trace-level attributes onto the root OTEL span before it is
  exported, without requiring any changes to graph code.

  ## Type of Change

  - [x] `feat`: New feature
  - [ ] `fix`: Bug fix
  - [ ] `ci`: CI/CD changes

  ## Related Issues

  Closes #213

  ## Changes Made

  - **`observability/span_enrichment.py`** (new file) — introduces `SpanEnrichmentProcessor` and
    `set_trace_context()`. The processor reads a `contextvars.ContextVar[dict[str, str] | None]` in
    `on_start()` and sets each stored key as a span attribute on the root span only
    (`span.parent is None`). Child spans (individual nodes, LLM calls, tool calls) are intentionally
    skipped — Langfuse reads `langfuse.user.id`, `langfuse.session.id`, and `langfuse.trace.name`
    exclusively from the root span, so enriching child spans would produce noise in per-observation
    metadata.

  - **`observability/otel.py`** — registers `SpanEnrichmentProcessor()` on the `TracerProvider`
    unconditionally in `setup()`, before any `BatchSpanProcessor` exporters. This guarantees
    attributes are present on the span before it is queued for export.

  - **`api/runs.py`** — calls `set_trace_context()` at the very top of `execute_run_async()`, which
    runs inside an isolated `asyncio.Task`. The context variable is task-local by design, so values
    set here are visible to all spans created during graph execution but never leak to concurrent
    requests.

  The following attributes are now set on every graph run's root span:

  | Attribute | Value | Where it appears |
  |---|---|---|
  | `langfuse.user.id` / `user.id` | authenticated user identity | User field in Langfuse; filterable in Phoenix |
  | `langfuse.session.id` / `session.id` | `thread_id` | Session field in Langfuse; filterable in Phoenix |
  | `langfuse.trace.name` | `graph_id` (e.g. `matter_agent`) | Trace name instead of "LangGraph" |
  | `langfuse.trace.metadata.run_id` | `run_id` | Filterable metadata in Langfuse |
  | `langfuse.trace.metadata.thread_id` | `thread_id` | Filterable metadata in Langfuse |
  | `langfuse.trace.metadata.graph_id` | `graph_id` | Filterable metadata in Langfuse |

  Both `langfuse.*` (Langfuse-native) and `user.id` / `session.id` (Phoenix/OpenInference) aliases
  are set so that a single `set_trace_context()` call works correctly regardless of which backend is
  configured in `OTEL_TARGETS`.

  ## Testing

  - [ ] E2E tests added/updated

  14 unit tests added in `tests/unit/test_observability/test_span_enrichment.py` covering:

  - `set_trace_context()` stores correct keys under both naming schemes
  - `set_trace_context()` with no arguments resets the context var to `None`
  - `metadata` dict keys are prefixed with `langfuse.trace.metadata.*`
  - `SpanEnrichmentProcessor.on_start()` sets attributes on root spans
  - `SpanEnrichmentProcessor.on_start()` skips child spans (`span.parent is not None`)
  - Context variable isolation between concurrent `asyncio.Task`s

  One existing test in `test_otel.py` updated: `test_setup_handles_target_errors_gracefully` —
  `add_span_processor` call count corrected from 1 to 2 to account for the unconditionally registered
  `SpanEnrichmentProcessor`.

  ## Checklist

  N/A

  ## Additional Notes

  `set_trace_context()` is intentionally public API — graph authors or custom middleware can call it
  to override or extend the default attributes set by Aegra. Example:

  ```python
  from aegra_api.observability.span_enrichment import set_trace_context

  set_trace_context(
      user_id=user.identity,
      session_id=thread_id,
      trace_name=graph_id,
      metadata={"tenant_id": tenant, "plan": "enterprise"},
  )
```

## Summary by CodeRabbit

* **New Features**
  * Per-request OpenTelemetry span enrichment for background executions: user ID, session ID, trace name, and custom metadata are attached to root spans and propagated into run execution.
  * Tracing pipeline now registers the enrichment processor so request context is captured prior to asynchronous graph execution.

* **Tests**
  * Added unit tests for trace-context setting, task-local isolation, and span enrichment behavior.

### DIFF
--- a/libs/aegra-api/src/aegra_api/api/runs.py
+++ b/libs/aegra-api/src/aegra_api/api/runs.py
@@ -25,7 +25,7 @@ from aegra_api.core.serializers import GeneralSerializer
 from aegra_api.core.sse import create_end_event, get_sse_headers
 from aegra_api.models import Run, RunCreate, RunStatus, User
 from aegra_api.models.errors import CONFLICT, NOT_FOUND, SSE_RESPONSE
-from aegra_api.observability.span_enrichment import set_trace_context
+from aegra_api.observability.span_enrichment import make_run_trace_context
 from aegra_api.services.broker import broker_manager
 from aegra_api.services.graph_streaming import stream_graph_events
 from aegra_api.services.langgraph_service import create_run_config, get_langgraph_service
@@ -262,8 +262,8 @@ async def create_run(
     # Build response from ORM -> Pydantic
     run = Run.model_validate(run_orm)
 
-    # Start execution asynchronously
-    # Don't pass the session to avoid transaction conflicts
+    # Start execution asynchronously.
+    # Don't pass the session to avoid transaction conflicts.
     task = asyncio.create_task(
         execute_run_async(
             run_id,
@@ -281,7 +281,8 @@ async def create_run(
             request.interrupt_after,
             request.multitask_strategy,
             request.stream_subgraphs,
-        )
+        ),
+        context=make_run_trace_context(run_id, thread_id, assistant.graph_id, user.identity),
     )
     logger.info(f"[create_run] background task created task_id={id(task)} for run_id={run_id}")
     active_runs[run_id] = task
@@ -383,8 +384,8 @@ async def create_and_stream_run(
     # Build response model for stream context
     run = Run.model_validate(run_orm)
 
-    # Start background execution that will populate the broker
-    # Don't pass the session to avoid transaction conflicts
+    # Start background execution that will populate the broker.
+    # Don't pass the session to avoid transaction conflicts.
     task = asyncio.create_task(
         execute_run_async(
             run_id,
@@ -402,7 +403,8 @@ async def create_and_stream_run(
             request.interrupt_after,
             request.multitask_strategy,
             request.stream_subgraphs,
-        )
+        ),
+        context=make_run_trace_context(run_id, thread_id, assistant.graph_id, user.identity),
     )
     logger.info(f"[create_and_stream_run] background task created task_id={id(task)} for run_id={run_id}")
     active_runs[run_id] = task
@@ -707,7 +709,7 @@ async def wait_for_run(
 
     # No pool connection held from here — safe for long waits
 
-    # Start execution asynchronously
+    # Start execution asynchronously.
     task = asyncio.create_task(
         execute_run_async(
             run_id,
@@ -725,7 +727,8 @@ async def wait_for_run(
             request.interrupt_after,
             request.multitask_strategy,
             request.stream_subgraphs,
-        )
+        ),
+        context=make_run_trace_context(run_id, thread_id, graph_id, user.identity),
     )
     logger.info(f"[wait_for_run] background task created task_id={id(task)} for run_id={run_id}")
     active_runs[run_id] = task
@@ -915,19 +918,6 @@ async def execute_run_async(
     subgraphs: bool | None = False,
 ) -> None:
     """Execute run asynchronously in background using streaming to capture all events."""
-    # Enrich OTEL spans with per-request trace context for this background task.
-    # Must be set before graph execution so SpanEnrichmentProcessor picks it up.
-    set_trace_context(
-        user_id=user.identity,
-        session_id=thread_id,
-        trace_name=graph_id,
-        metadata={
-            "run_id": run_id,
-            "thread_id": thread_id,
-            "graph_id": graph_id,
-        },
-    )
-
     owns_session = session is None
     if session is None:
         maker = _get_session_maker()
--- a/libs/aegra-api/src/aegra_api/observability/span_enrichment.py
+++ b/libs/aegra-api/src/aegra_api/observability/span_enrichment.py
@@ -27,7 +27,7 @@ from opentelemetry.sdk.trace import ReadableSpan, Span, SpanProcessor
 
 # Per-request context variable holding span attributes to inject.
 # None means no trace context is set; on_start() is a no-op in that case.
-_trace_attrs: contextvars.ContextVar[dict[str, str] | None] = contextvars.ContextVar(
+_trace_attrs: contextvars.ContextVar[dict[str, str | int | float | bool] | None] = contextvars.ContextVar(
     "aegra_otel_trace_attrs", default=None
 )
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: libs/aegra-api/src/aegra_api/observability/span_enrichment.py [hunk 1]
```diff
@@ -0,0 +1,117 @@
...
+    per-observation metadata.
+
+    Call :func:`set_trace_context` inside the asyncio Task that runs
+    graph execution to populate the context variable before any spans
+    are created.
+    """
+
+    def on_start(self, span: Span, parent_context: Context | None = None) -> None:
+        if span.parent is not None:
+            # Only enrich root spans — child spans don't need trace-level attrs.
```
ibbybuilds: **Bug:** If an incoming request carries a W3C `traceparent` header (e.g., from a frontend, API gateway, or another service in a distributed trace), the OTEL SDK will set `span.parent` to a **remote** `SpanContext`. This means the local root span of your request will have `span.parent is not None`, and this check will skip it — no enrichment happens.

The fix is to only skip **local** child spans:

```suggestion
        if span.parent is not None and span.parent.is_valid and not span.parent.is_remote:
            # Only enrich root spans — child spans don't need trace-level attrs.
            return
```

This way, spans with a remote parent (the local root of a distributed trace) still get enriched, which is exactly what Langfuse expects.

libs/aegra-api/src/aegra_api/observability/span_enrichment.py
```diff
@@ -0,0 +1,117 @@
...
+    def force_flush(self, timeout_millis: int = 30000) -> bool:
+        return True
+
+
+def set_trace_context(
+    *,
+    user_id: str | None = None,
+    session_id: str | None = None,
+    trace_name: str | None = None,
+    metadata: dict[str, str] | None = None,
```
ibbybuilds: **Nit:** This type is narrower than what OTEL attributes support. Metadata values could reasonably be `int`, `float`, or `bool` — all valid OTEL attribute types. Consider:

```suggestion
    metadata: dict[str, str | int | float | bool] | None = None,
```

This avoids forcing callers to stringify numeric values (e.g., retry counts, latencies).

libs/aegra-api/src/aegra_api/observability/span_enrichment.py
```diff
@@ -0,0 +1,117 @@
...
+            (Phoenix).
+        trace_name: Human-readable trace name (typically the graph ID).
+            Sets ``langfuse.trace.name``.
+        metadata: Arbitrary key/value pairs to attach as filterable
+            metadata.  Each key is stored as
+            ``langfuse.trace.metadata.<key>`` so that Langfuse exposes
+            it as a queryable field (rather than burying it under
+            ``metadata.attributes``).  Phoenix already receives these
+            values via ``cfg["metadata"]`` through LangChainInstrumentor,
+            so no duplicate path is needed there.
```
ibbybuilds: Same here — should match the parameter type:

```suggestion
    attrs: dict[str, str | int | float | bool] = {}
```

libs/aegra-api/src/aegra_api/api/runs.py
```diff
@@ -910,6 +915,19 @@ async def execute_run_async(
...
+        user_id=user.identity,
+        session_id=thread_id,
+        trace_name=graph_id,
+        metadata={
+            "run_id": run_id,
+            "thread_id": thread_id,
+            "graph_id": graph_id,
+        },
+    )
+
```
ibbybuilds: **Fragility concern:** This relies on `execute_run_async` always being called via `asyncio.create_task()` for context isolation. If it's ever called directly (or via `await`), the context leaks to the caller's scope and could affect concurrent requests.

Two options (pick whichever fits better):

1. **Defensive `copy_context()`** — wrap the call site that spawns this task:
   ```python
   ctx = contextvars.copy_context()
   task = asyncio.get_event_loop().create_task(ctx.run(execute_run_async, ...))
   ```

2. **Document the contract** — add a comment at the call site that creates the task, making the requirement explicit so future refactors don't break it.

Not blocking, but worth hardening.

libs/aegra-api/src/aegra_api/observability/span_enrichment.py
```diff
@@ -0,0 +1,117 @@
...
+            ``langfuse.session.id`` (Langfuse) and ``session.id``
+            (Phoenix).
+        trace_name: Human-readable trace name (typically the graph ID).
+            Sets ``langfuse.trace.name``.
+        metadata: Arbitrary key/value pairs to attach as filterable
+            metadata.  Each key is stored as
+            ``langfuse.trace.metadata.<key>`` so that Langfuse exposes
+            it as a queryable field (rather than burying it under
+            ``metadata.attributes``).  Phoenix already receives these
+            values via ``cfg["metadata"]`` through LangChainInstrumentor,
```
ibbybuilds: **Nit:** This docstring references "LangChainInstrumentor" — should be removed or reworded per project conventions.

PR #250, opened 2026-03-18
Feat/base image overlay pattern

andreassiegel:
## Description

Split the monolithic Dockerfile into a reusable base image (Aegra core only) and a thin overlay pattern for consumer projects to add their custom graphs and dependencies. The existing root `docker-compose.yml` continues to work unchanged with example agents via volume mounts.

## Type of Change

- [x] `feat`: New feature
- [ ] `fix`: Bug fix
- [x] `docs`: Documentation changes
- [ ] `ci`: CI/CD changes

## Related Issues

None

## Changes Made

- **`deployments/docker/Dockerfile`** — Install both `aegra-api` and `aegra-cli` from workspace, remove `aegra.json`/`examples/` from image, add `libpq5` to runtime stage, add OCI labels
- **`deployments/docker/Dockerfile.overlay`** (new) — Thin overlay that installs custom deps via `uv` and copies project `aegra.json` + `src/`
- **`deployments/docker/docker-compose.overlay.yml`** (new) — Reference compose for consumer projects (PostgreSQL + overlay service with health checks)
- **`docker-compose.yml`** — Add header comments clarifying dev vs overlay usage
- **`docs/guides/deployment.mdx`** — New "Base image + overlay pattern" section with build instructions and project structure guide
- **`README.md`** — Update deployment feature description

## Testing

- [ ] E2E tests added/updated

Manual test steps:
1. Built base image: `docker build -f deployments/docker/Dockerfile -t aegra-base:test .`
2. Verified CLI available: `docker run --rm aegra-base:test aegra version`
3. Verified `pip list | grep aegra` shows both packages installed
4. Root `docker compose up --build -d` starts successfully with example agents via volume mount

## Checklist

- [x] Commit messages follow [[Conventional Commits](https://www.conventionalcommits.org/)](https://www.conventionalcommits.org/)

N/A

## Additional Notes

- The base image uses `--no-dev` and `--no-emit-workspace` flags during dependency export to keep only production dependencies
- Consumer projects control their base image version via the `AEGRA_BASE_IMAGE` build arg
- The overlay Dockerfile includes build tools (`build-essential`, `libpq-dev`) for native extensions and cleans them up after install to minimize image size

## Summary by CodeRabbit

* **New Features**
  * Introduced base image overlay pattern for flexible Docker deployments
  * Added docker-compose overlay configuration with PostgreSQL integration

* **Documentation**
  * Updated deployment guide with base image overlay pattern instructions
  * Enhanced README with new deployment approach documentation

* **Improvements**
  * Optimized Docker builds with multi-stage architecture
  * Enhanced development environment setup with source code and configuration mounting

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ aegra version           # Show version info
 | [Streaming](https://docs.aegra.dev/guides/streaming) | 8 SSE stream modes with reconnection |
 | [Store](https://docs.aegra.dev/guides/store) | Key-value and semantic search storage |
 | [Observability](https://docs.aegra.dev/guides/observability) | Fan-out tracing to Langfuse, Phoenix, or any OTLP backend |
-| [Deployment](https://docs.aegra.dev/guides/deployment) | Docker, PaaS, and Kubernetes deployment |
+| [Deployment](https://docs.aegra.dev/guides/deployment) | Docker, PaaS, Kubernetes, and base image overlay pattern |
 | [Migration](https://docs.aegra.dev/migration) | Migrate from LangSmith Deployments |
 
 ## 💬 Community & Support
--- a/deployments/docker/Dockerfile
+++ b/deployments/docker/Dockerfile
@@ -1,9 +1,9 @@
 # =============================================================================
 # Aegra Base Image
 # =============================================================================
-# Reusable base image containing only the Aegra core runtime + Alembic
-# migrations. Does NOT include aegra.json or agent code — those are added
-# by consumer projects via an overlay Dockerfile or volume mounts.
+# Reusable base image containing the Aegra core runtime (aegra-api + aegra-cli)
+# and Alembic migrations. Does NOT include aegra.json or agent code — those are
+# added by consumer projects via an overlay Dockerfile or volume mounts.
 #
 # Build:
 #   docker build -f deployments/docker/Dockerfile -t aegra-base:0.8.2 .
@@ -43,22 +43,31 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     libpq-dev \
     && rm -rf /var/lib/apt/lists/*
 
-# Copy aegra-api package files and workspace lockfile
-COPY libs/aegra-api/pyproject.toml libs/aegra-api/README.md ./
-COPY uv.lock ./
+# Copy workspace definition + both package manifests + lockfile.
+# This allows uv to resolve the full dependency tree for aegra-cli
+# (which transitively includes all aegra-api dependencies).
+COPY pyproject.toml uv.lock ./
+COPY libs/aegra-api/pyproject.toml libs/aegra-api/README.md ./libs/aegra-api/
+COPY libs/aegra-cli/pyproject.toml libs/aegra-cli/README.md ./libs/aegra-cli/
 
-# Install production dependencies only (no dev deps like langchain-openai, pytest, etc.).
-RUN uv export --frozen --no-emit-project --no-dev --format=requirements-txt > requirements.txt && \
+# Install all external production dependencies for aegra-cli (includes
+# aegra-api deps transitively: fastapi, langgraph, sqlalchemy, ...
+# plus CLI deps: click, rich, jinja2, python-dotenv).
+# --no-emit-workspace excludes workspace packages themselves (installed from source below).
+RUN uv export --frozen --package aegra-cli --no-emit-project --no-emit-workspace --no-dev \
+        --format=requirements-txt > requirements.txt && \
     uv pip install --system --compile-bytecode -r requirements.txt && \
     rm requirements.txt
 
-# Copy the actual project source code and forced includes (alembic).
-COPY libs/aegra-api/src/ ./src/
-COPY libs/aegra-api/alembic.ini ./alembic.ini
-COPY libs/aegra-api/alembic/ ./alembic/
+# Copy source code for both packages.
+COPY libs/aegra-api/src/ ./libs/aegra-api/src/
+COPY libs/aegra-api/alembic.ini ./libs/aegra-api/alembic.ini
+COPY libs/aegra-api/alembic/ ./libs/aegra-api/alembic/
+COPY libs/aegra-cli/src/ ./libs/aegra-cli/src/
 
-# Install the project package itself.
-RUN uv pip install --system --compile-bytecode --no-deps .
+# Install both workspace packages from local source (no deps — already installed above).
+RUN uv pip install --system --compile-bytecode --no-deps ./libs/aegra-api && \
+    uv pip install --system --compile-bytecode --no-deps ./libs/aegra-cli
 
 # -----------------------------
 # Final, minimal runtime image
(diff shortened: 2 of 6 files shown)


ibbybuilds (commented):
## Deep Code Review — PR #250

**`Feat/base image overlay pattern`**

Reviewed 6 files, +261/-18 lines. The base+overlay split is the right approach and well-documented. Several issues need attention before merge.

---

### Blockers

#### 1. Version hardcoded at wrong value — `0.8.2` should be `0.8.3`
**Files:** `Dockerfile.overlay:26`, `docker-compose.overlay.yml:43`, `Dockerfile` header comment

`aegra-base:0.8.2` appears as the default in three places, but both `pyproject.toml` files show version `0.8.3`. A consumer who copies `Dockerfile.overlay` without overriding `AEGRA_BASE_IMAGE` will reference a base image that may not exist or does not match their aegra-api version.

**Fix:** Change all occurrences of `0.8.2` to `0.8.3`. Also document the version-sync requirement in the release checklist (CLAUDE.md versioning section).

#### 2. `COPY pyproject.toml uv.loc[k] ./` contradicts `uv export --frozen`
**File:** `Dockerfile.overlay:44,48`

The Docker glob `uv.loc[k]` makes the lockfile optional (COPY succeeds with no file). But `uv export --frozen` requires a lockfile and will fail if absent. These are contradictory — the glob gives false reassurance that the lockfile is optional.

**Fix (pick one):**
- Make lockfile mandatory: `COPY pyproject.toml uv.lock ./` (straightforward, matches docs which say "run `uv lock` to generate")
- Or add a guard: `RUN test -f uv.lock || (echo "ERROR: uv.lock required. Run 'uv lock' locally." && exit 1)`

---

### Good to fix

#### 3. `libpq-dev` not purged in overlay cleanup
**File:** `Dockerfile.overlay:55-58`

`apt-get purge --auto-remove build-essential` does NOT remove `libpq-dev` — it was explicitly installed, not a dependency of `build-essential`. Dev headers (~500KB) remain in the final image.

```diff
- apt-get purge -y --auto-remove build-essential && \
+ apt-get purge -y --auto-remove build-essential libpq-dev && \
```

#### 4. `libpq5` comment is misleading — `psycopg[binary]` ships its own libpq
**File:** `Dockerfile:52`

Comment says `# libpq5 needed by psycopg` but `aegra-api` pins `psycopg[binary]>=3.2.9`, which bundles a statically-linked libpq. The system `libpq5` is not actually used. Either remove it (saves ~1MB) or fix the comment:

```diff
- # Install only minimal runtime libs (libpq5 needed by psycopg)
+ # Install only minimal runtime libs
+ # Note: libpq5 is a safety net if psycopg is switched from [binary] to [c] extra
```

#### 5. Default password in reference compose
**File:** `docker-compose.overlay.yml:20`

`POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-aegra_secret}` — consumers who copy this file verbatim will ship with a well-known password.

**Fix:** Add a prominent comment:
```yaml
# WARNING: Override POSTGRES_PASSWORD in .env before deploying to any non-local environment
POSTGRES_PASSWORD: ${POSTGRES_PASSWORD:-change_me_in_dot_env}
```

#### 6. Transitive dependency chain is fragile and undocumented
**File:** `Dockerfile:50`

`uv export --package aegra-cli --no-emit-workspace` works because aegra-api deps flow transitively through aegra-cli's `aegra-api~=0.8.0` dependency. If that constraint is ever relaxed or removed, the base image silently loses API runtime deps.

**Fix:** Add a comment explaining the chain:
```dockerfile
# aegra-api deps are captured transitively via aegra-cli's dependency on aegra-api.
# --no-emit-workspace excludes workspace packages (installed from source below).
```

#### 7. Dev compose missing alembic volume mount
**File:** `docker-compose.yml`

Source and config are volume-mounted but `alembic/` is not. A developer who creates a new migration after building won't see it in the container (hot-reload only covers Python source).

**Fix:** Add `./libs/aegra-api/alembic:/app/alembic:ro` to the volumes block.

#### 8. `USER app` — document why CLI is in the base image
**File:** `Dockerfile` builder stage

The CLI is needed for the `CMD ["aegra", "serve"]` entrypoint. This is correct but a consumer might wonder why a CLI tool is baked into their production API image. A one-line comment would clarify:

```dockerfile
# Install both packages: aegra-cli provides the "aegra serve" entrypoint command
RUN uv pip install --system --compile-bytecode --no-deps ./libs/aegra-api && \
    uv pip install --system --compile-bytecode --no-deps ./libs/aegra-cli
```

---

### Impact Analysis

- **Blast radius:** Low — Docker/deployment infrastructure only, no application code changes
- **Breaking changes:** No — root `docker-compose.yml` continues to work unchanged
- **Performance:** Base image is now leaner (no examples/agent code baked in)
- **Consumer impact:** Overlay pattern provides clear separation for downstream projects

### Architectural Notes
- **Pattern consistency:** Good — standard Docker base+overlay pattern
- **Documentation:** Good — deployment guide updated with clear instructions
- **User assumption (USER app):** Verified — base image creates the `app` user at line 14
- **Build reproducibility:** Warning — the `uv.loc[k]` glob undermines `--frozen` guarantee
- **Version synchronization:** Warning — no mechanism to keep `0.8.x` in Dockerfiles in sync with `pyproject.toml` versions

JiwaniZakir (commented):
In `Dockerfile.overlay`, the `COPY pyproject.toml uv.loc[k] ./` glob trick is intended to make `uv.lock` optional, but the `uv export --frozen` command on the very next `RUN` step will immediately fail if no lockfile is present — making the optionality effectively meaningless. Either drop the `--frozen` flag and document that unlocked resolution is acceptable for overlay consumers, or require the lockfile explicitly with `COPY pyproject.toml uv.lock ./` and update the comment accordingly.

Additionally, the overlay's `uv export` call omits `--no-emit-workspace` (present in the base `Dockerfile`'s equivalent step), which could cause problems if a consumer project happens to use a uv workspace layout — the workspace packages themselves would be emitted into `requirements.txt` and pip would try to resolve them as external packages. Adding `--no-emit-workspace` defensively here would make the behavior consistent with the base image build.

ibbybuilds:
@andreassiegel thanks for the pr, the base+overlay split is the right pattern and the docs were well thought out. i am going to close this though because after looking at how it fits with `aegra init` i think the integration point is wrong.

`aegra init` already generates a standalone Dockerfile and docker-compose for every new project. adding a second set of reference Dockerfiles in `deployments/docker/` creates two competing answers to "how do i dockerize my aegra project" and users would have to discover the overlay pattern separately.

the right path forward is:
1. publish an official `aegra-base` image to ghcr as part of the release workflow
2. add a `--base-image` flag to `aegra init` that generates the thin overlay Dockerfile instead of the standalone one
3. keep the current standalone Dockerfile as the default for new users

this way the overlay pattern is discoverable through the same cli flow everyone already uses, not a separate directory they have to find.

your review feedback (version hardcoding, uv.lock/frozen contradiction, libpq-dev cleanup, default password) is all captured and will carry forward into the implementation. opening a tracking issue now.

appreciate the contribution, the pattern itself is solid and we will be using it.

PR #259, opened 2026-03-25
fix: correct wrong ThreadHistoryRequest signature for before

Jourdelune:
## Description
Fix the `before` parameter type signature in `ThreadHistoryRequest` to accept `RunnableConfig` objects instead of plain strings. This resolves HTTP 422 validation errors when using LangChain's official `useStream()` SDK for paginating thread history.

The issue occurs because LangChain's `useStream()` sends pagination cursors as Config objects (`{ configurable: { checkpoint_id: "..." } }`), but aegra's Pydantic model expected a plain string. LangGraph's `aget_state_history()` also expects a `RunnableConfig`, not a string.

## Type of Change
- [x] `fix`: Bug fix

## Changes Made
- Changed `before: str | None` to `before: dict | None` in `ThreadHistoryRequest` model (`aegra_api/models/threads.py`)

## Testing
- Verified pagination works with LangChain's `useStream()` hook
- Confirmed HTTP 422 validation error is resolved

## Additional Notes
This fix ensures aegra correctly implements the LangGraph Server API specification for the `/threads/{thread_id}/history` endpoint. The `before` parameter should match LangGraph SDK's `RunnableConfig` type.

## Summary by CodeRabbit

* **Refactor**
  * Updated API request parameter structure to support object-based data formats instead of string-based formats, enabling more flexible request handling.

### DIFF
--- a/libs/aegra-api/src/aegra_api/models/threads.py
+++ b/libs/aegra-api/src/aegra_api/models/threads.py
@@ -145,7 +145,7 @@ class ThreadHistoryRequest(BaseModel):
     """Request model for thread history endpoint"""
 
     limit: int | None = Field(10, ge=1, le=1000, description="Number of states to return")
-    before: str | None = Field(None, description="Return states before this checkpoint ID")
+    before: dict | None = Field(None, description="Return states before this checkpoint ID")
     metadata: dict[str, Any] | None = Field(None, description="Filter by metadata")
     checkpoint: dict[str, Any] | None = Field(None, description="Checkpoint for subgraph filtering")
     subgraphs: bool | None = Field(False, description="Include states from subgraphs")


JiwaniZakir (commented):
The field description in `ThreadHistoryRequest.before` still reads `"Return states before this checkpoint ID"` — "checkpoint ID" implies a string identifier, but the type is now `dict`. That mismatch will confuse API consumers trying to understand what structure to pass. The description should be updated to document the expected dict schema (e.g., which keys are required or optional).

This is also a breaking change: any existing callers passing a plain string checkpoint ID will now fail Pydantic validation silently at the boundary rather than getting a meaningful error. It's worth considering whether `str | dict | None` is the safer migration path, or at minimum adding a `@field_validator` that rejects unexpected shapes so the failure mode is explicit.

Finally, there's no type narrowing on the `dict` — using `dict[str, Any]` (consistent with the `metadata` and `checkpoint` fields just below it) would at least align the style and make static analysis more useful across the model.

PR #258, opened 2026-03-24
feat: add LangGraph.js support via Node.js bridge

fouadmahmoud283-ai:
## Description

Add end-to-end LangGraph.js support to Aegra, enabling users to run TypeScript-based LangGraph.js graphs alongside Python graphs on the same server.

## Type of Change

- [x] feat — New feature

## Related Issues

Closes #N/A — New feature request

### Architecture: Node.js Bridge (JSON-RPC over stdin/stdout)
- **`libs/aegra-js-bridge/`** — TypeScript package that runs as a subprocess, loads `@langchain/langgraph` graphs, and exposes them via JSON-RPC 2.0 protocol
- **`services/js_bridge.py`** — Async Python client managing the Node.js subprocess lifecycle (spawn, health checks, restart, graceful shutdown)
- **`services/js_graph_wrapper.py`** — Pregel-compatible wrapper so JS graphs work transparently with Aegra's existing streaming, checkpointing, and run machinery

### Core Integration
- **`langgraph_service.py`** — Updated `_load_langgraphjs_graph()` to use the built-in bridge (with fallback to external `langgraph_api.js` loader for backward compatibility)
- **`graph_streaming.py`** — Added `JSGraphWrapper` detection for JS graph streaming alongside existing `BaseRemotePregel` support

### CLI Integration
- **`templates/__init__.py`** — Registered new "LangGraph.js Chatbot" template (option 3 in `aegra init`)
- **`commands/init.py`** — Generates JS-specific `aegra.json` with `"runtime": "langgraphjs"` object format
- **`cli.py`** — `aegra dev` auto-detects Node.js, checks version (18+), and runs `npm install` for JS graph directories

### Examples & Documentation
- **`examples/langgraphjs_chatbot/`** — Working TypeScript chatbot example (graph.ts, state.ts, package.json)
- **`docs/guides/langgraphjs.mdx`** — Complete setup, configuration, architecture, and troubleshooting guide
- **`docs/reference/configuration.mdx`** — Updated with built-in bridge description
- **`docs/feature-support.mdx`** — Added LangGraph.js to feature matrix

### Key Design Decisions
- **JSON-RPC over stdin/stdout** — Self-contained, no extra ports needed
- **Checkpointing on Python side** — JS bridge stays stateless, PostgreSQL persistence managed by Aegra
- **Static graphs only** for initial release (factory functions can be added later)
- **`tsx`** for direct TypeScript loading without a build step

## Testing

[...]

PR #261, opened 2026-03-28
Add A2A and MCP Adapters to mirror LangSmith Platform

shadowcodex:
## Description
Adds two new "adapters" to the server. 

- MCP
- A2A

Based on the implementation and functionality provided by LangSmith Platform today, the goal was to create a simple layer that reuses existing API methods and authentication. This allows adapters to follow any changes in how certain processes work under the hood.

This is paired with some tests that also evaluate them using MCP and A2A client libraries.

## Type of Change

- [x] `feat`: New feature
- [ ] `fix`: Bug fix
- [x] `docs`: Documentation changes
- [x] `test`: Tests added/updated
- [ ] `ci`: CI/CD changes

## Related Issues
https://github.com/ibbybuilds/aegra/issues/75

## Changes Made

- New adapter pattern (Adapter + Service) for adding interfaces for other common protocols, expandable to add more later.
- Main Updated to init and mount routes/services for MCP & A2A
- A2A Service and Adapter Added
- MCP Service and Adapter Added
- Docs added and updated where needed for it.

## Testing

- [x] E2E tests added/updated

## Checklist

- [ ] Type checking passes (`make type-check`) 
- [ ] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)
- [ ] PR title follows Conventional Commits format

## Additional Notes

Type Check Failed but seems like many things are also failing that which are existing.

## Summary by CodeRabbit

* **New Features**
  * A2A agent-to-agent JSON-RPC endpoints, agent discovery, and streaming messaging; MCP endpoint exposing agents as tools (streamable HTTP). Adapters enabled by default and respect existing auth.

* **Configuration**
  * New http flags to enable/disable A2A and MCP adapters.

* **Documentation**
  * New A2A and MCP guides, README updates, and config examples with client snippets.

* **Tests**
  * Extensive unit, integration, and end-to-end tests covering A2A/MCP and auth.

* **Chores**
  * Package version bumps.

### DIFF
--- a/libs/aegra-api/src/aegra_api/adapters/a2a_adapter.py
+++ b/libs/aegra-api/src/aegra_api/adapters/a2a_adapter.py
@@ -5,6 +5,7 @@ discovery at /.well-known/agent-card.json.
 Uses the `a2a-sdk` library.
 """
 
+from collections.abc import Awaitable, Callable
 from typing import Any
 
 import structlog
@@ -58,7 +59,7 @@ def _rpc_error(code: int, message: str, rpc_id: Any) -> dict[str, Any]:
 # ---------------------------------------------------------------------------
 
 
-def _make_well_known_handler(service: A2AService):  # type: ignore[return]
+def _make_well_known_handler(service: A2AService) -> Callable[..., Awaitable[JSONResponse]]:
     """Return a route handler for GET /.well-known/agent-card.json.
 
     Args:
@@ -121,7 +122,7 @@ def _make_well_known_handler(service: A2AService):  # type: ignore[return]
     return _well_known_agent_card
 
 
-def _make_agent_cards_handler(service: A2AService):  # type: ignore[return]
+def _make_agent_cards_handler(service: A2AService) -> Callable[..., Awaitable[JSONResponse]]:
     """Return a route handler for GET /a2a/agent-cards.
 
     Args:
@@ -167,7 +168,7 @@ def _make_agent_cards_handler(service: A2AService):  # type: ignore[return]
     return _list_agent_cards
 
 
-def _make_rpc_handler(service: A2AService):  # type: ignore[return]
+def _make_rpc_handler(service: A2AService) -> Callable[..., Awaitable[Response]]:
     """Return a route handler for POST /a2a/{assistant_id}.
 
     Args:
@@ -229,7 +230,7 @@ def _make_rpc_handler(service: A2AService):  # type: ignore[return]
                 logger.error(
                     "A2A send_message error", error=str(exc), assistant_id=assistant_id
                 )
-                return JSONResponse(content=_rpc_error(-32603, f"Internal error: {exc}", rpc_id))
+                return JSONResponse(content=_rpc_error(-32603, "Internal server error", rpc_id))
             return JSONResponse(content=_rpc_ok(result, rpc_id))
 
         if method == "tasks/get":
@@ -241,12 +242,12 @@ def _make_rpc_handler(service: A2AService):  # type: ignore[return]
                     content=_rpc_error(-32602, "Missing required param: id", rpc_id)
                 )
             try:
-                task_result: dict[str, Any] = await service.get_task(task_id_param)
+                task_result: dict[str, Any] = await service.get_task(task_id_param, user)
             except ValueError as exc:
                 return JSONResponse(content=_rpc_error(-32602, str(exc), rpc_id))
             except Exception as exc:
                 logger.error("A2A get_task error", error=str(exc), task_id=task_id_param)
-                return JSONResponse(content=_rpc_error(-32603, f"Internal error: {exc}", rpc_id))
+                return JSONResponse(content=_rpc_error(-32603, "Internal server error", rpc_id))
             return JSONResponse(content=_rpc_ok(task_result, rpc_id))
 
         if method == "tasks/cancel":
@@ -259,12 +260,12 @@ def _make_rpc_handler(service: A2AService):  # type: ignore[return]
                 )
             try:
                 await streaming_service.cancel_run(cancel_task_id)
-                cancelled_task: dict[str, Any] = await service.get_task(cancel_task_id)
+                cancelled_task: dict[str, Any] = await service.get_task(cancel_task_id, user)
             except ValueError as exc:
                 return JSONResponse(content=_rpc_error(-32602, str(exc), rpc_id))
             except Exception as exc:
                 logger.error("A2A tasks/cancel error", error=str(exc), task_id=cancel_task_id)
-                return JSONResponse(content=_rpc_error(-32603, f"Internal error: {exc}", rpc_id))
+                return JSONResponse(content=_rpc_error(-32603, "Internal server error", rpc_id))
             return JSONResponse(content=_rpc_ok(cancelled_task, rpc_id))
 
         if method == "message/stream":
@@ -295,7 +296,7 @@ def _make_rpc_handler(service: A2AService):  # type: ignore[return]
                 logger.error(
                     "A2A message/stream error", error=str(exc), assistant_id=assistant_id
                 )
-                return JSONResponse(content=_rpc_error(-32603, f"Internal error: {exc}", rpc_id))
+                return JSONResponse(content=_rpc_error(-32603, "Internal server error", rpc_id))
 
         return JSONResponse(
             content=_rpc_error(-32601, f"Method not found: {method!r}", rpc_id)
@@ -326,6 +327,9 @@ def mount_a2a(app: FastAPI) -> None:
     """
     service: A2AService = get_a2a_service()
 
+    # Discovery endpoints are intentionally unauthenticated per the A2A spec.
+    # Agent cards are public metadata (graph IDs, capabilities) — not sensitive.
+    # This matches LangSmith's behavior.
     app.add_api_route(
         "/.well-known/agent-card.json",
         _make_well_known_handler(service),
--- a/libs/aegra-api/src/aegra_api/adapters/a2a_adapter.py
+++ b/libs/aegra-api/src/aegra_api/adapters/a2a_adapter.py
@@ -1,0 +1,353 @@
+"""A2A (Agent-to-Agent) protocol adapter.
+
+Exposes agents via A2A JSON-RPC at /a2a/{assistant_id} with agent card
+discovery at /.well-known/agent-card.json.
+Uses the `a2a-sdk` library.
+"""
+
+from typing import Any
+
+import structlog
+from fastapi import Depends, FastAPI, HTTPException, Request
+from fastapi.responses import JSONResponse, StreamingResponse
+from sqlalchemy.ext.asyncio import AsyncSession
+from starlette.responses import Response
+
+from aegra_api.adapters.a2a_service import A2AService, get_a2a_service
+from aegra_api.core.auth_deps import get_current_user
+from aegra_api.core.orm import get_session
+from aegra_api.models.auth import User
+from aegra_api.services.streaming_service import streaming_service
+
+logger = structlog.get_logger(__name__)
+
+# ---------------------------------------------------------------------------
+# JSON-RPC helpers
+# ---------------------------------------------------------------------------
+
+
+def _rpc_ok(result: Any, rpc_id: Any) -> dict[str, Any]:
+    """Build a JSON-RPC 2.0 success response.
+
+    Args:
+        result: The result payload to include.
+        rpc_id: The id from the original request.
+
+    Returns:
+        JSON-RPC success response dict.
+    """
+    return {"jsonrpc": "2.0", "result": result, "id": rpc_id}
+
+
+def _rpc_error(code: int, message: str, rpc_id: Any) -> dict[str, Any]:
+    """Build a JSON-RPC 2.0 error response.
+
+    Args:
+        code: The JSON-RPC error code.
+        message: Human-readable error description.
+        rpc_id: The id from the original request.
+
+    Returns:
+        JSON-RPC error response dict.
+    """
+    return {"jsonrpc": "2.0", "error": {"code": code, "message": message}, "id": rpc_id}
+
+
+# ---------------------------------------------------------------------------
+# Route handler factories (closures capture the service instance)
+# ---------------------------------------------------------------------------
+
+
+def _make_well_known_handler(service: A2AService):  # type: ignore[return]
+    """Return a route handler for GET /.well-known/agent-card.json.
+
+    Args:
+        service: The A2AService instance to use for building cards.
+
+    Returns:
+        An async route handler function.
+    """
+
+    async def _well_known_agent_card(
+        request: Request, assistant_id: str | None = None
+    ) -> JSONResponse:
+        """Return the agent card for the given assistant or the first registered agent.
+
+        Args:
+            request: The incoming FastAPI request (used to derive base URL).
+            assistant_id: Optional query param to select a specific agent.
+
+        Returns:
+            JSONResponse with the agent card dict.
+
+        Raises:
+            HTTPException: 404 when no agents are registered or assistant_id is unknown.
+        """
+        registry: dict[str, Any] = (
+            service._langgraph_service._graph_registry if service._langgraph_service else {}
+        )
+
+        if not registry:
+            raise HTTPException(status_code=404, detail="No agents registered")
+
+        if assistant_id is not None:
+            if assistant_id not in registry:
+                raise HTTPException(
+                    status_code=404, detail=f"Unknown assistant_id: {assistant_id!r}"
+                )
+            target_id = assistant_id
+        else:
+            target_id = next(iter(registry))
+
+        graph_meta: Any = registry[target_id]
+        name: str = (
+            graph_meta.get("name", target_id) if isinstance(graph_meta, dict) else target_id
+        )
+        description: str = (
+            graph_meta.get("description", f"Agent: {target_id}")
+            if isinstance(graph_meta, dict)
+            else f"Agent: {target_id}"
+        )
+
+        base_url: str = str(request.base_url).rstrip("/")
+        card: dict[str, Any] = service.build_agent_card(
+            assistant_id=target_id,
+            name=name,
+            description=description,
+            base_url=base_url,
+        )
+        return JSONResponse(content=card)
+
+    return _well_known_agent_card
+
+
+def _make_agent_cards_handler(service: A2AService):  # type: ignore[return]
+    """Return a route handler for GET /a2a/agent-cards.
+
+    Args:
+        service: The A2AService instance to use for building cards.
+
+    Returns:
+        An async route handler function.
+    """
+
+    async def _list_agent_cards(request: Request) -> JSONResponse:
+        """Return an array of agent cards for all registered agents.
+
+        Args:
+            request: The incoming FastAPI request (used to derive base URL).
+
+        Returns:
+            JSONResponse with a JSON array of agent card dicts.
+        """
+        registry: dict[str, Any] = (
+            service._langgraph_service._graph_registry if service._langgraph_service else {}
+        )
+
+        base_url: str = str(request.base_url).rstrip("/")
+        cards: list[dict[str, Any]] = []
+        for agent_id, graph_meta in registry.items():
+            name: str = (
+                graph_meta.get("name", agent_id) if isinstance(graph_meta, dict) else agent_id
+            )
+            description: str = (
+                graph_meta.get("description", f"Agent: {agent_id}")
+                if isinstance(graph_meta, dict)
+                else f"Agent: {agent_id}"
+            )
+            card: dict[str, Any] = service.build_agent_card(
+                assistant_id=agent_id,
+                name=name,
+                description=description,
+                base_url=base_url,
+            )
+            cards.append(card)
+        return JSONResponse(content=cards)
+
+    return _list_agent_cards
+
+
+def _make_rpc_handler(service: A2AService):  # type: ignore[return]
+    """Return a route handler for POST /a2a/{assistant_id}.
+
+    Args:
+        service: The A2AService instance to dispatch JSON-RPC calls to.
+
+    Returns:
+        An async route handler function.
+    """
+
+    async def _a2a_rpc(
+        request: Request,
+        assistant_id: str,
+        user: User = Depends(get_current_user),
+        session: AsyncSession = Depends(get_session),
+    ) -> Response:
+        """Handle A2A JSON-RPC requests for a specific agent.
+
+        Parses the JSON-RPC body and dispatches to the appropriate A2AService
+        method based on the ``method`` field.
+
+        Supported methods:
+        - ``message/send``: Execute the agent and return the resulting Task.
+        - ``tasks/get``: Look up an existing task by ID.
+        - ``tasks/cancel``: Cancel a running task and return its updated state.
+        - ``message/stream``: Not yet implemented; returns -32603 error.
+
+        Args:
+            request: The incoming FastAPI request.
+            assistant_id: Path parameter identifying the target agent.
+
+        Returns:
+            JSONResponse containing a JSON-RPC 2.0 response envelope.
+        """
+        try:
+            body: dict[str, Any] = await request.json()
+        except Exception:
+            return JSONResponse(content=_rpc_error(-32700, "Parse error: invalid JSON", None))
+
+        rpc_id: Any = body.get("id")
+        method: str = body.get("method", "")
+        params: dict[str, Any] = body.get("params") or {}
+
+        if method == "message/send":
+            message: dict[str, Any] = params.get("message") or {}
+            parts: list[dict[str, Any]] = message.get("parts") or []
+            context_id: str | None = params.get("contextId") or message.get("contextId")
+            task_id: str | None = params.get("taskId") or message.get("taskId")
+            try:
+                result: dict[str, Any] = await service.send_message(
+                    assistant_id=assistant_id,
+                    parts=parts,
+                    user=user,
+                    context_id=context_id,
+                    task_id=task_id,
+                )
+            except ValueError as exc:
+                return JSONResponse(content=_rpc_error(-32602, str(exc), rpc_id))
+            except Exception as exc:
+                logger.error(
+                    "A2A send_message error", error=str(exc), assistant_id=assistant_id
+                )
+                return JSONResponse(content=_rpc_error(-32603, f"Internal error: {exc}", rpc_id))
+            return JSONResponse(content=_rpc_ok(result, rpc_id))
+
+        if method == "tasks/get":
+            task_id_param: str | None = (
+                params.get("id") or params.get("taskId")
+            ) if params else None
+            if not task_id_param:
+                return JSONResponse(
+                    content=_rpc_error(-32602, "Missing required param: id", rpc_id)
+                )
+            try:
+                task_result: dict[str, Any] = await service.get_task(task_id_param)
+            except ValueError as exc:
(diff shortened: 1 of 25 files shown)


ibbybuilds (commented):
## Deep Code Review — PR #261

**`Add A2A and MCP Adapters`**

Reviewed 27 files, +3443/-5 lines. Solid adapter pattern and good test coverage at all levels. Several issues should be addressed before merge.

---

### Blockers

#### 1. `break` in `stream_message` only exits the inner `for` loop, not the outer `async for`
**File:** `a2a_service.py:351-353` — Confidence: 97

When `event: end` is seen, the `break` exits the inner `for line in chunk.strip().split("\n")` loop but the outer `async for chunk in streaming_response.body_iterator` keeps iterating, consuming further chunks and potentially yielding duplicate artifact events.

```python
# Fix: add a sentinel flag
stream_ended = False
async for chunk in streaming_response.body_iterator:
    if stream_ended:
        break
    ...
    for line in chunk.strip().split("\n"):
        ...
        elif line.startswith("event: end"):
            stream_ended = True
            break
```

#### 2. Exception details leaked in every JSON-RPC error response (CWE-209)
**File:** `a2a_adapter.py:232,249,267,298` — Confidence: 97

Every `except Exception` branch passes the raw exception into the response: `f"Internal error: {exc}"`. This can expose SQL fragments, file paths, and internal library details to callers. The full exception is already logged.

**Fix:** Replace with `"Internal server error"`. Same for MCP auth middleware (`mcp_adapter.py:63-67`): replace `f"Authentication failed: {exc}"` with `"Authentication failed"`.

#### 3. No ownership check on `get_task` / `tasks/cancel` — any user can read/cancel any task (CWE-639)
**File:** `a2a_service.py:378-417` — Confidence: 96

`get_task` looks up a run solely by `run_id` with no check that the run belongs to the requesting user. Any authenticated user who knows another user's task_id can read its output or cancel it.

**Fix:** Add `user: User` parameter and filter: `RunORM.user_id == user.identity`. Update all call sites in `a2a_adapter.py` to pass `user`.

#### 4. `send_message` hardcodes `TaskState.completed` — failed runs reported as success
**File:** `a2a_service.py:215-233` — Confidence: 95

If the graph run ends with an error status but `wait_for_run` does not raise, the A2A Task is unconditionally returned with `TaskState.completed`. The actual run status is never checked.

#### 5. `taskId` returned from `send_message` is disconnected from the DB `run_id`
**File:** `a2a_service.py:247` — Confidence: 95

`effective_task_id = task_id or str(uuid4())` — but `wait_for_run` internally creates its own `run_id`. The caller receives a task ID that does not match any DB record, so `tasks/get` will always return "Task not found" for tasks created via `message/send`. The two-step `send -> get` flow is broken by design.

**Fix:** Either have `wait_for_run` accept a caller-provided `run_id`, or return the created `run_id` so the adapter can use it as the `taskId`.

---

### Good to fix

#### 6. MCP `_AuthMiddleware` returns 401 when auth backend returns `None`
**File:** `mcp_adapter.py:68-73` — Confidence: 92

When the auth backend has no `_authenticate_handler` configured, it returns `None` (meaning "skip auth, allow through"). But the MCP middleware treats `None` as a 401 rejection, blocking all MCP requests when auth is loaded without a handler.

**Fix:** Pass through instead of rejecting when result is `None`.

#### 7. `convert_output_to_parts` — `isinstance(msg, AIMessage)` fails silently on dict messages
**File:** `a2a_service.py:106-115` — Confidence: 90

LangGraph may return messages as plain dicts. The `isinstance` check will skip every dict message, returning empty artifacts with no error.

**Fix:** Add `elif isinstance(msg, dict) and msg.get("type") == "ai":` branch.

#### 8. No-op ternary — dead branch
**File:** `a2a_service.py:196,289` — Confidence: 99

```python
graph_id = assistant_id if assistant_id in registry else assistant_id
```
Both branches produce the same value. Simplify to `graph_id = assistant_id`.

#### 9. Both services access private `_graph_registry` and `_get_base_graph`
**Files:** `a2a_service.py`, `mcp_service.py`, `a2a_adapter.py` — Confidence: 85

Both adapter services bypass the public API of `LangGraphService`. If internals are refactored, adapters break silently.

**Suggestion:** Add public methods to `LangGraphService`:
- `list_graph_ids() -> list[str]`
- `get_graph_schema(graph_id: str) -> dict[str, Any]`

#### 10. A2A calls route handler functions directly (`wait_for_run`, `create_and_stream_run`)
**File:** `a2a_service.py` — Confidence: 85

These are FastAPI route handlers, not service-layer functions. `stream_message` even consumes `StreamingResponse.body_iterator` directly — an ASGI implementation detail. Consider extracting core execution logic into a service function that both the route handlers and adapters can call.

#### 11. `_langgraph_service: Any` — should be typed
**Files:** `a2a_service.py:132`, `mcp_service.py:28`

Use `LangGraphService | None` instead of `Any`. If this causes a circular import, use `from __future__ import annotations`.

#### 12. Discovery endpoints expose full graph registry unauthenticated (CWE-200)
**File:** `a2a_adapter.py:61-167`

`/.well-known/agent-card.json` and `/a2a/agent-cards` enumerate every graph ID with no auth. If intentional per A2A spec, add a comment documenting this. If graph IDs are sensitive, consider requiring auth.

#### 13. `# type: ignore[return]` on handler factories — missing return type annotations
**File:** `a2a_adapter.py:61,124,170`

Annotate the return type of the factory functions instead of suppressing.

#### 14. Importing private functions across modules
**Files:** `mcp_adapter.py:27` (`_to_user_model`), `a2a_service.py:40` (`_get_session_maker`)

Promote these to public APIs or restructure to avoid cross-module private access.

#### 15. `disable_a2a` / `disable_mcp` — inconsistent naming
**File:** `config.py`

Existing convention uses `enable_custom_route_auth` (positive). New flags use `disable_X` (negative). Consider `enable_a2a: true` / `enable_mcp: true` for consistency.

#### 16. Hard dependencies on early-stage SDKs
**File:** `pyproject.toml`

`mcp>=1.26.0` and `a2a-sdk[http-server]>=0.3.25` are installed for every deployment even when disabled. Consider:
- Upper pin: `mcp>=1.26.0,<2.0.0`
- Optional extras: `aegra-api[a2a,mcp]`
- Add `langchain-core` as explicit dep (currently implicit via `langgraph`)

---

### Impact Analysis

- **Blast radius:** Medium — new code only, no changes to existing endpoints
- **Breaking changes:** No — purely additive
- **New dependencies:** `mcp>=1.26.0`, `a2a-sdk[http-server]>=0.3.25` (always installed)
- **Performance:** Startup cost for loading adapter modules; per-request cost is delegated to existing services

### Architectural Notes
- **Pattern consistency:** Good — Service+adapter split is clean and follows project conventions
- **Test coverage:** Good — Unit, integration, and E2E tests at all levels
- **Documentation:** Good — Docs added for both A2A and MCP
- **Coupling:** Warning — Direct access to `LangGraphService` private members; calling route handlers as service functions; consuming `StreamingResponse.body_iterator`
- **Auth consistency:** Warning — A2A uses FastAPI `Depends`, MCP uses ASGI middleware (justified by architecture, but `_to_user_model` import is fragile)

hukaixuan:
Just came across this PR — it looks very relevant to our use case. We have an urgent need for A2A capabilities, so it would be awesome to see this merged and released soon if it’s ready.
Thanks so much for your work on this!

shadowcodex:
Hey @hukaixuan , its waiting for @ibbybuilds to review and suggest changes.

Hew as going to look at it after the worker queue redis stuff so maybe he can take a look next week.

ibbybuilds:
hey Shannon, sorry this sat for a month. read through it today. want to align before line-level review.

architecture is solid. adapter pattern translates inbound → RunCreate → existing runs handlers. reuses auth, threads, persistence, worker dispatch. plays fine with worker mode and the langgraph.js bridge in #258. future protocols (telegram, slack, grpc) plug in the same way.

things i'd want before merge:

1. auth pattern is split. a2a uses fastapi `Depends(get_current_user)`, mcp uses a custom asgi middleware + contextvar. two different paths to maintain and to reason about for future adapters. if it can be unified on one pattern that's preferable. if mcp genuinely needs the asgi route because tool handlers don't see the request, fine, but call that out in `adapters/__init__.py` so future adapter authors pick the right one.

2. mcp auth middleware lets the request through when `result is None`. auth_deps for the runs path raises 401 in the same case. minor inconsistency, but for an endpoint that exposes every graph as a tool i'd rather match the stricter side.

3. a2a-sdk pin is stale. `>=0.3.25` while 1.0.2 is current stable. bump to `>=1.0.0,<2.0.0`. add upper bound on mcp too: `>=1.26.0,<2.0.0`.

4. rebase against main. there are conflicts and the version in pyproject is behind, lint and check-commits failing too.

dropping inline comments on the line-level stuff i found. push back on anything that feels off.

libs/aegra-api/src/aegra_api/adapters/a2a_adapter.py
```diff
@@ -0,0 +1,357 @@
...
+        if method == "tasks/cancel":
+            cancel_task_id: str | None = (
+                params.get("id") or params.get("taskId")
+            ) if params else None
+            if not cancel_task_id:
+                return JSONResponse(
+                    content=_rpc_error(-32602, "Missing required param: id", rpc_id)
+                )
+            try:
+                await streaming_service.cancel_run(cancel_task_id)
```
ibbybuilds, line 262: auth bypass — `streaming_service.cancel_run(cancel_task_id)` runs before any ownership check, so any authed user can cancel another user's run by guessing the run id. `service.get_task(cancel_task_id, user)` does check ownership but only after the cancel already happened.

flip the order so we verify ownership first, then cancel:

```python
try:
    await service.get_task(cancel_task_id, user)  # raises ValueError if not owned
except ValueError as exc:
    return JSONResponse(content=_rpc_error(-32602, str(exc), rpc_id))
await streaming_service.cancel_run(cancel_task_id)
cancelled_task = await service.get_task(cancel_task_id, user)
```

libs/aegra-api/src/aegra_api/adapters/a2a_service.py
```diff
@@ -0,0 +1,459 @@
...
+        effective_task_id = task_id or str(uuid4())
+
+        request = RunCreate(
+            assistant_id=resolved_assistant_id,
+            input=input_data,
+            stream_mode=["messages"],
+        )
+
+        # Get the streaming response from the runs API
+        streaming_response = await create_and_stream_run(thread_id, request, user, session)
```
ibbybuilds, line 324: the injected fastapi `session` is held open for the entire sse stream lifetime here, can be minutes, which starves the connection pool. `wait_for_run` solves the same problem by using `_get_session_maker()` directly, drop the param and do the same:

```python
async def stream_message(
    self,
    *,
    assistant_id: str,
    parts: list[dict[str, Any]],
    user: User,
    context_id: str | None = None,
    task_id: str | None = None,
) -> AsyncGenerator[str, None]:
    ...
    maker = _get_session_maker()
    async with maker() as session:
        streaming_response = await create_and_stream_run(thread_id, request, user, session)
    # session released, stream now proceeds without holding a pool connection
```

then drop `session=session` from the adapter call site too.

libs/aegra-api/src/aegra_api/adapters/a2a_adapter.py
```diff
@@ -0,0 +1,357 @@
...
+                        "Connection": "keep-alive",
+                    },
+                )
+            except ValueError as exc:
+                return JSONResponse(content=_rpc_error(-32602, str(exc), rpc_id))
+            except Exception as exc:
+                logger.error(
+                    "A2A message/stream error", error=str(exc), assistant_id=assistant_id
+                )
+                return JSONResponse(content=_rpc_error(-32603, "Internal server error", rpc_id))
```
ibbybuilds, line 299: `service.stream_message(...)` returns the async generator object instantly, none of its body runs at call time. so `ValueError` from `convert_parts_to_langchain` and `RuntimeError` from the uninitialized service check never reach this try/except, they fire on the first iteration inside `StreamingResponse` after the 200 + headers have already gone out. clients see a mid-stream tcp cut.

wrap iteration in an inner generator that converts errors into a final `TaskStatusUpdateEvent(state=failed)`:

```python
async def _guarded_stream():
    try:
        async for chunk in service.stream_message(
            assistant_id=assistant_id, parts=parts, user=user,
            context_id=context_id, task_id=task_id,
        ):
            yield chunk
    except ValueError as exc:
        logger.warning("A2A stream invalid params", error=str(exc))
        err = TaskStatusUpdateEvent(
            taskId=task_id or str(uuid4()),
            contextId=context_id or str(uuid4()),
            status=TaskStatus(state=TaskState.failed),
            final=True,
        )
        yield f"data: {err.model_dump_json(by_alias=True)}\n\n"
    except Exception as exc:
        logger.error("A2A stream error", error=str(exc))
        err = TaskStatusUpdateEvent(
            taskId=task_id or str(uuid4()),
            contextId=context_id or str(uuid4()),
            status=TaskStatus(state=TaskState.failed),
            final=True,
        )
        yield f"data: {err.model_dump_json(by_alias=True)}\n\n"

return StreamingResponse(_guarded_stream(), media_type="text/event-stream", headers={...})
```

libs/aegra-api/src/aegra_api/adapters/a2a_service.py
```diff
@@ -0,0 +1,459 @@
...
+                                        artifact=Artifact(
+                                            artifactId="output",
+                                            parts=[Part(root=TextPart(text=content))],
+                                        ),
+                                        append=True,
+                                        lastChunk=False,
+                                    )
+                                    yield f"data: {artifact_event.model_dump_json(by_alias=True)}\n\n"
+
+                elif line.startswith("event: end"):
```
ibbybuilds, line 373: the upstream streaming service emits `event: error` when a run fails (see `streaming_service.py` `broker.put(end_event_id, ("end", {"status": "error"}))`). the parser only checks `event: end` here, so failed runs fall through and the loop ends up emitting `TaskState.completed` unconditionally below. a2a clients can't tell success from failure.

```python
elif line.startswith("event: end"):
    stream_ended = True
    break
elif line.startswith("event: error"):
    stream_failed = True
    stream_ended = True
    break
```

then map at the end: `TaskState.failed if stream_failed else TaskState.completed`.

libs/aegra-api/src/aegra_api/adapters/a2a_adapter.py
```diff
@@ -0,0 +1,357 @@
...
+        include_in_schema=False,
+    )
+    # /a2a/agent-cards MUST be registered before /a2a/{assistant_id} so the
+    # static path wins over the dynamic one.
+    app.add_api_route(
+        "/a2a/agent-cards",
+        _make_agent_cards_handler(service),
+        methods=["GET"],
+        include_in_schema=False,
+    )
```
ibbybuilds, line 346: `/.well-known/agent-card.json` and `/a2a/agent-cards` are unauthenticated by design here. they expose the full graph id list and capabilities, which in a self-hosted deployment is often business-logic naming (`crm-pipeline`, `hr-lookup`, etc). every other route in aegra requires auth, so this is new unauth surface.

a2a spec only requires the well-known endpoint be reachable, it doesn't forbid auth. add an opt-in flag, default-deny:

```python
deps = [] if (http_config or {}).get("public_agent_cards", False) else [Depends(get_current_user)]
app.add_api_route("/.well-known/agent-card.json", ..., dependencies=deps)
app.add_api_route("/a2a/agent-cards", ..., dependencies=deps)
```

libs/aegra-api/src/aegra_api/adapters/a2a_adapter.py
```diff
@@ -0,0 +1,357 @@
...
+            try:
+                result: dict[str, Any] = await service.send_message(
+                    assistant_id=assistant_id,
+                    parts=parts,
+                    user=user,
+                    context_id=context_id,
+                    task_id=task_id,
+                )
+            except ValueError as exc:
+                return JSONResponse(content=_rpc_error(-32602, str(exc), rpc_id))
```
ibbybuilds, line 228: `str(exc)` is forwarded verbatim to the client. the service raises with `!r` formatting, e.g. `Task not found: 'abc-123'` and `Unknown tool: 'crm-pipeline'`, which gives an attacker an oracle to confirm run ids and enumerate the graph registry by triggering `ValueError` paths.

log the detail, return a generic message:

```python
except ValueError as exc:
    logger.warning("a2a invalid params", error=str(exc), method=method)
    return JSONResponse(content=_rpc_error(-32602, "Invalid parameters", rpc_id))
```

same pattern applies at lines 247, 265, 294.

libs/aegra-api/src/aegra_api/adapters/a2a_adapter.py
```diff
@@ -0,0 +1,357 @@
...
+
+        Returns:
+            JSONResponse containing a JSON-RPC 2.0 response envelope.
+        """
+        try:
+            body: dict[str, Any] = await request.json()
+        except Exception:
+            return JSONResponse(content=_rpc_error(-32700, "Parse error: invalid JSON", None))
+
+        rpc_id: Any = body.get("id")
```
ibbybuilds, line 210: json-rpc 2.0 spec restricts `id` to string, number, or null. `Any` here lets a caller pass an arbitrary nested object that gets reflected back in every response (and in error responses too). easy amplification primitive.

```python
raw_id = body.get("id")
rpc_id: str | int | None = raw_id if isinstance(raw_id, (str, int, type(None))) else None
```

libs/aegra-api/src/aegra_api/adapters/a2a_service.py
```diff
@@ -0,0 +1,459 @@
...
+        registry: dict[str, Any] = self._langgraph_service._graph_registry
+
+        graph_id = assistant_id
+        resolved_assistant_id = resolve_assistant_id(graph_id, registry)
+
+        # Convert parts to LangChain messages
+        langchain_messages = convert_parts_to_langchain(parts)
+        input_data: dict[str, Any] = {"messages": langchain_messages}
+
+        thread_id = context_id if context_id is not None else str(uuid4())
```
ibbybuilds, line 233: `context_id` (and `task_id` below) come straight from json-rpc params and get used as `thread_id` / `run_id`, stored in the db, returned in responses. no length or format check, so a 10mb string sails through. orm prevents sql injection but not garbage storage.

```python
import re
_UUID_RE = re.compile(r"^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$", re.I)

if context_id is not None and not _UUID_RE.match(context_id):
    raise ValueError("contextId must be a valid UUID")
```

apply the same to `task_id`.

libs/aegra-api/src/aegra_api/adapters/mcp_adapter.py
```diff
@@ -0,0 +1,168 @@
...
+    a 401 JSON response.
+    """
+
+    def __init__(self, app: ASGIApp) -> None:
+        self._app = app
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send) -> None:
+        if scope["type"] != "http":
+            await self._app(scope, receive, send)
+            return
```
ibbybuilds, line 55: non-http scopes skip the contextvar set + auth check entirely. lifespan is fine, but other scope types (websocket if mcp ever adds it, or anything unexpected) reach `self._app` unauthenticated. tighten to allowlist:

```python
if scope["type"] == "lifespan":
    await self._app(scope, receive, send)
    return
if scope["type"] != "http":
    response = StarletteJSONResponse(status_code=400, content={"error": "Unsupported scope type"})
    await response(scope, receive, send)
    return
```

libs/aegra-api/src/aegra_api/adapters/mcp_service.py
```diff
@@ -0,0 +1,123 @@
...
+                graph = await self._langgraph_service._get_base_graph(graph_id)
+                input_schema = graph.get_input_jsonschema()
+            except Exception:
+                logger.exception("mcp_list_tools_graph_load_failed", graph_id=graph_id)
+                continue
+
+            tools.append(
+                {
+                    "name": graph_id,
+                    "description": f"Run the {graph_id} agent",
```
ibbybuilds, line 68: description is hardcoded as `Run the {graph_id} agent`, but `docs/mcp.md` says "the graph's description if set, otherwise a default description". either read a `description` field from the graph config dict and use it here, or update the doc to match. doc currently lies.

libs/aegra-api/src/aegra_api/adapters/a2a_adapter.py
```diff
@@ -0,0 +1,357 @@
...
+            request: The incoming FastAPI request.
+            assistant_id: Path parameter identifying the target agent.
+
+        Returns:
+            JSONResponse containing a JSON-RPC 2.0 response envelope.
+        """
+        try:
+            body: dict[str, Any] = await request.json()
+        except Exception:
+            return JSONResponse(content=_rpc_error(-32700, "Parse error: invalid JSON", None))
```
ibbybuilds, line 208: `except Exception:` here swallows every error with no log. for json parsing the right type is `json.JSONDecodeError` (or `UnicodeDecodeError` for bad encoding):

```python
except (json.JSONDecodeError, UnicodeDecodeError):
    return JSONResponse(content=_rpc_error(-32700, "Parse error: invalid JSON", None))
```

ibbybuilds (changes requested):
Please check comments.

ibbybuilds:
two design suggestions, not blockers, just wanted to surface them while it's fresh.

**composite task ids**: right now task_id is the raw run_id, which is what makes the cancel ownership check awkward (you have to look the run up to know which thread it belongs to). encoding it as `{context_id}:{run_id}` and parsing back when needed makes tasks/get and tasks/cancel auth checks one call instead of two, and removes the path where you cancel before verifying ownership. a2a spec doesn't require uuid task ids, any string works.

**a2a-sdk weight**: the sdk pulls in protobuf, googleapis-common-protos, json-rpc, google-api-core. heavy for what we use it for here (mostly TaskState enum, TaskStatusUpdateEvent, TaskArtifactUpdateEvent shapes). a typeddict mirror of the wire format would drop the install footprint a lot. tradeoff is we have to track spec changes ourselves. fine either way, raising in case the dep cost matters.

ibbybuilds:
Shannon, three months of silence from me after you had already been waiting a month. Sorry, that is not how I want to run this.

Where things actually stand, so you are not guessing.

On MCP: #383 also adds an MCP adapter, and both PRs create adapters/mcp_adapter.py, touch config.py, and mount /mcp. Only one MCP implementation can land. My current lean is #383 for that half, mainly because it uses the standalone fastmcp package rather than the bundled one and has a cleaner auth story. I have not made that final.

On A2A: this is the part I actually want from your PR, and nothing else in the repo provides it. @hukaixuan asked for A2A on this thread back in April and I still owe them an answer too.

I am not asking you to change anything right now. Adapters are parked this cycle behind a security fix and a batch of bug fixes, and I am not going to ask you to delete working code on the promise that a different PR might land later. Once #383 is either merged or dropped I will come back here with a concrete ask, and if the answer is A2A only then that is a smaller and better scoped PR than what you are carrying now.

The four points from my April review still stand for the A2A half whenever we pick this back up, particularly the a2a-sdk pin and the composite task id suggestion for the cancel ownership check.

Thanks for waiting this long, and for telling hukaixuan honestly that it was on me.

shadowcodex:
@ibbybuilds I think I need the A2A scope more, so I'm happy to rework this to just be A2A adapter.

m-hisham08:
@shadowcodex @ibbybuilds

Asking as a user rather than a reviewer.

We run Aegra in production and now need to expose our agent to a third-party client agent over A2A. 

We’re trying to work out whether to plan around this landing or write our own adapter in the meantime.

Is A2A support still on the cards here? Is there any rough sense of timeline?

Completely fine if the answer is “no timeline”. It just helps to know which way to plan.

bcrowther-spaceheater:
> @shadowcodex @ibbybuilds
> 
> Asking as a user rather than a reviewer.
> 
> We run Aegra in production and now need to expose our agent to a third-party client agent over A2A.
> 
> We’re trying to work out whether to plan around this landing or write our own adapter in the meantime.
> 
> Is A2A support still on the cards here? Is there any rough sense of timeline?
> 
> Completely fine if the answer is “no timeline”. It just helps to know which way to plan.

Also wondering this?

ibbybuilds:
@shadowcodex  are you still willing to work on A2A?

m-hisham08:
Hey @ibbybuilds, since there hasn't been any follow-up from @shadowcodex, would it be alright if I picked this up? Happy to take a stab at it if that'd be useful.

shadowcodex:
@m-hisham08 please do. I am focused on hardware implementations atm and not langgraph servers. Would be great if you could take this to finish line.

ibbybuilds:
@m-hisham08 go ahead, and thanks for picking this up.

I specced the whole A2A wire contract at #625 before replying. It is long, but it is verified rather than inferred: I ran the Python client (`a2a-sdk` 1.1.5) against a capture server, read the JavaScript client (`@a2a-js/sdk` 1.2.0) at its tag, and checked both against the A2A conformance suite. Every "observed" in it is captured bytes. Better to hand you that once than trickle corrections over three review rounds.

**What is already right in the PR and should stay:**

- Mounted at `/a2a/{assistant_id}`, matching Platform. The docs table in the PR says `/a2a`; the table is stale, the routing is correct.
- Reuses the existing run create and join paths instead of forking them.
- `http.disable_a2a`, matching config key.
- Around 1000 lines of tests.

**Read §0 first.** Our agent card decides which protocol dialect every client speaks to us. Both SDKs read `protocolVersion` off the card; absent, it reads as 0.3.0 and selects slash method names with lowercase enums. With `protocolVersion: "1.0"` plus a non-empty `supportedInterfaces`, they switch to PascalCase and `TASK_STATE_*`. The two are mutually incompatible in both directions.

That reframes where the PR stands. It is not that the modern names are missing; it is that a current client cannot talk to it at all, and the card is where that gets fixed.

**Three items decide whether a modern client works. In this order:**

1. **The card** (§6). `protocolVersion: "1.0"`, non-empty `supportedInterfaces`, non-empty `skills`, built as a plain dict rather than through `a2a-sdk` model classes. Also `capabilities.streaming: true` must be present: the Python SDK silently takes the non-streaming path when it is absent, whatever the caller configured.

2. **The two dialects** (§3). Accept both method families, answer each in its own shape. The required-field asymmetry table is the part that bites: legacy requires `status`, `contextId` and `final`; modern requires none of them but rejects unknown fields outright.

3. **Modern part shapes on input** (§7). Modern sends flat parts with no `kind`: `{"text": "hi"}`, `{"raw": ..., "mediaType": ...}`. Legacy nests under `file` with `mimeType`. Every conformance fixture uses the flat form, so a `kind`-discriminated validator rejects all of them.

After those, **§10 (resume promotion from parts)** is the largest single piece: three ways a client can resume an interrupt, with different strictness and an ordering that matters.

One counterintuitive correction worth calling out: **`final: true` does not end a stream.** I held a socket open after sending it and the Python client blocked for the full six seconds, returning only at close. It never reads the field. We have to close the connection. §11 covers the rest, including `data: ` with the space and one JSON object per line, both of which the conformance parser is strict about.

The remaining nine items are additive and can land in separate PRs if you would rather keep them small. `SubscribeToTask` and the push-notification methods are explicitly out of scope for the first pass.

MCP in this PR is a separate question. I have not specced it, and I would rather not merge a protocol surface we have not pinned down, for the reason above. Happy to split MCP into its own PR so A2A is not held up behind it, whichever you prefer.

Shout if anything in the spec is unclear or looks wrong. Some of it documents behavior that reads like a bug, and where that happens the reasoning for matching it anyway is written next to it.

PR #322, opened 2026-04-21
fix(api): honor sort params and metadata types in /threads/search

### LINKED ISSUE #302: [BUG] Thread search not working
**Describe the bug**  
Thread search is not working correctly when filters are applied. Search without filters may return results, but once filters are added, the thread search does not return the expected threads or behaves incorrectly.

**To Reproduce**  
1. Set up the environment with Aegra configured and running  
2. Create or use existing threads with searchable content  
3. Call the thread search endpoint with one or more filters applied  
4. Observe that the results are empty, incorrect, or do not match the filters  

**Expected behavior**  
Thread search should return the correct threads matching the provided filters.

**Actual behavior**  
When filters are applied, thread search fails to return the expected results or returns incorrect data.

**Environment:**  
- OS: [e.g. macOS 14.5, Ubuntu 22.04]  
- Python version: [e.g. 3.11.5]  
- Aegra version: [e.g. 0.1.0]  
- LangGraph version: [e.g. 0.6.4]  

**Configuration:**  
- Database: [e.g. PostgreSQL 15]  
- Auth type: [e.g. noop, jwt]  
- Deployment: [e.g. local, docker, kubernetes]  

**Logs/Error Messages:**  
return [] when filters applied

arturbagramyan1:
## Description

`POST /threads/search` had three related defects that made sorting and metadata filtering unreliable. This PR fixes all three without changing the response shape or breaking existing callers.

1. **`order_by` was schema-only.** `ThreadSearchRequest` declared `order_by: str = "created_at DESC"` but the handler hardcoded `ORDER BY created_at DESC` and never read the field. Ascending was impossible.
2. **langgraph-sdk's native sort shape was dropped.** `client.threads.search(...)` sends `sort_by: ThreadSortBy` and `sort_order: "asc" | "desc"` as separate top-level fields; aegra didn't declare them, so Pydantic stripped them silently.
3. **Metadata filter broke for non-string scalars.** The filter compiled to `(metadata_json->>'key') = :value` with `str(value)` bound — so `metadata={"active": True}` sent `"True"` (capital T), mismatching both JSON `true` and JSON string `"true"` (Postgres `->>` returns lowercase). Any `bool` / `int` / `float` / `None` filter returned zero rows.

Three commits, one per bug, each revertable independently.

## Type of Change

- [ ] `feat`: New feature
- [x] `fix`: Bug fix
- [ ] `ci`: CI/CD changes

(Commit 2 is `feat(api):` on its own because it adds the two new request fields, but the PR-level intent is fix.)

## Related Issues

Fixes #302 — "[BUG] Thread search not working" (filters returning `[]`). That issue's symptom is the Bug 3 metadata-filter type mismatch described above; Bugs 1 and 2 are adjacent defects surfaced during the same investigation.

## Changes Made

- **`libs/aegra-api/src/aegra_api/api/threads.py`** — added `_resolve_sort(request)` helper that whitelists sort columns (`created_at`, `updated_at`, `thread_id`, `status`) and returns `(column, is_ascending)`. Applied in `search_threads` instead of the hardcoded `ThreadORM.created_at.desc()`. Replaced `ThreadORM.metadata_json[key].as_string() == str(value)` with `ThreadORM.metadata_json[key] == cast(bindparam(None, value=value, type_=JSON), JSONB)` so equality respects JSONB types.
- **`libs/aegra-api/src/aegra_api/models/threads.py`** — added `sort_by: str | None` and `sort_order: Literal["asc", "desc"] | None` to `ThreadSearchRequest`. `order_by` kept for backward compatibility with its original default.
- **Precedence:** `sort_by` > `order_by` > default (`created_at DESC`). Unknown columns, SQL-injection attempts, empty strings, and `state_updated_at` (present in the SDK literal but absent from aegra's schema) all fall back silently to default — never 500.
- **`sort_order`** is a strict `Literal`; bad values return 422. `sort_by` is loose (silent fallback) because the SDK may add new literals we don't yet support.
- No writer changes, no migration. An earlier hypothesis was that writers were coercing booleans to strings; reading the code showed writers always store metadata as-is. The bug lived entirely on the read side.

## Testing

- [x] Unit tests added/updated — `libs/aegra-api/tests/unit/test_api/test_threads_search_sort.py` (15 cases: default, ASC/DESC parsing, case-insensitivity, SQL-injection fallback, empty string, SDK shape, default-desc when only `sort_by` is set, precedence, `state_updated_at` fallback, unknown `sort_by` fallback).
- [x] Integration tests added/updated — extended `TestSearchThreads` in `libs/aegra-api/tests/integration/test_api/test_threads_crud.py` with 6 new cases covering 200-on-ASC, no-500-on-malformed, SDK shape 200, `state_updated_at` 200, invalid `sort_order` returns 422, and bool-metadata request body acceptance.
- [x] E2E tests added/updated — new `libs/aegra-api/tests/e2e/test_threads/test_search.py` (6 cases against real Postgres: ASC, DESC, SDK shape, precedence, bool metadata filter returning the correct typed subset, malformed fallback).
- [x] Manual testing performed — full stack via `docker compose up -d`, curled all three bug scenarios via `/threads/search` and verified OpenAPI schema at `/docs` exposes the new fields.

### Test run

```
libs/aegra-api/tests/unit + integration: 1185 passed in 16.49s
libs/aegra-api/tests/e2e/test_threads/test_search.py: 6 passed in 6.81s
```

## Checklist

- [x] Code follows project style (Ruff formatting) — `uv run ruff check` clean on all modified files
- [x] Type checking passes (`make type-check`) — `uv run ty check` unchanged vs. `main` (pre-existing diagnostics only, none on my additions)
- [ ] Documentation updated (if needed) — endpoint docs at `/docs` auto-update from the Pydantic model. No prose docs referenced `order_by` or this endpoint's body shape specifically; happy to add a note if a guide page should be updated.

## Additional Notes

**Commits (intentionally three, each revertable independently):**

- `b8aedc8` `fix(api): apply order_by in threads search`
- `f0f17ff` `feat(api): accept sort_by/sort_order in threads search (SDK compat)`
- `d9faffa` `fix(api): match metadata boolean filters via JSONB cast`

**Backward compatibility:**

- `order_by` still works exactly as advertised — callers using the old default `"created_at DESC"` are unaffected.
- `sort_by` / `sort_order` are additive.
- Metadata filters that were always sent with string values (status enums, names, tags, UUIDs) behave identically.
- Filters sending `{"active": "true"}` (string) as a workaround for the old bug will now only match rows where `active` was literally stored as the string `"true"` — the correct JSONB-typed behavior. Callers should send real booleans.

**Not done (deliberately):**

- No writer-side coercion of booleans. Writers were not the cause; callers control what they persist.
- No backfill migration. The bug is on the read side; existing rows don't need rewriting.
- No changes to `list_threads` (`GET /threads`), no changes to `/threads/search` response shape, no new query params besides `sort_by` / `sort_order`.

## Summary by CodeRabbit

* **New Features**
  - Thread search accepts SDK-style sorting parameters (`sort_by`, `sort_order`) and uses them when provided.

* **Bug Fixes**
  - Metadata filtering now matches JSON values correctly (booleans, numbers, nulls).
  - Malformed/unknown sort inputs fall back safely without errors.

* **Improvements**
  - Stable pagination for tied sort keys via a deterministic secondary sort (prevents duplicates).
  - Legacy `order_by` kept but deprecated in favor of `sort_by` + `sort_order`.

* **Database**
  - Added a GIN index on thread metadata to speed containment-based JSON queries.

* **Tests**
  - New unit, integration, and E2E tests for sorting, precedence, metadata filtering, and pagination.

### DIFF
--- a/libs/aegra-api/src/aegra_api/api/threads.py
+++ b/libs/aegra-api/src/aegra_api/api/threads.py
@@ -3,16 +3,15 @@
 import asyncio
 import contextlib
 import json
+import warnings
 from datetime import UTC, datetime
 from typing import Any
 from uuid import uuid4
 
 import structlog
 from fastapi import APIRouter, Depends, HTTPException, Query
-from sqlalchemy import bindparam, cast, select
-from sqlalchemy.dialects.postgresql import JSONB
+from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncSession
-from sqlalchemy.types import JSON
 
 from aegra_api.core.active_runs import active_runs
 from aegra_api.core.auth_deps import auth_dependency, get_current_user
@@ -54,25 +53,30 @@ _DEFAULT_SORT_ASC = False
 def _resolve_sort(request: ThreadSearchRequest) -> tuple[Any, bool]:
     """Resolve (ORM column, is_ascending) for /threads/search.
 
-    Precedence: SDK-style ``sort_by`` (with optional ``sort_order``) wins over the
-    legacy ``order_by`` string. Unknown columns or malformed input silently fall
-    back to the default (``created_at DESC``) — never 500.
+    Precedence: SDK-style ``sort_by`` (Pydantic-validated against the column
+    Literal) wins over the legacy ``order_by`` string. ``order_by`` stays
+    permissive — unknown columns or malformed input silently fall back to the
+    default (``created_at DESC``) — to preserve backward compatibility.
     """
-    field = _DEFAULT_SORT_FIELD
-    asc = _DEFAULT_SORT_ASC
-
     if request.sort_by:
-        candidate = request.sort_by.strip().lower()
-        if candidate in _ALLOWED_SORT_FIELDS:
-            field = candidate
-            asc = (request.sort_order or "desc").lower() == "asc"
-    elif request.order_by:
-        parts = request.order_by.strip().split()
-        if parts and parts[0].lower() in _ALLOWED_SORT_FIELDS:
-            field = parts[0].lower()
-            asc = len(parts) > 1 and parts[1].lower() == "asc"
+        column = getattr(ThreadORM, request.sort_by)
+        asc = (request.sort_order or "desc").lower() == "asc"
+        return column, asc
 
-    return getattr(ThreadORM, field), asc
+    # order_by is marked deprecated on the Pydantic Field for OpenAPI; the
+    # access-site warning is meant for callers, not for the handler honouring
+    # the field — suppress it here.
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", DeprecationWarning)
+        legacy = request.order_by
+
+    if legacy:
+        parts = legacy.strip().split()
+        if parts and parts[0].lower() in _ALLOWED_SORT_FIELDS:
+            asc = len(parts) > 1 and parts[1].lower() == "asc"
+            return getattr(ThreadORM, parts[0].lower()), asc
+
+    return getattr(ThreadORM, _DEFAULT_SORT_FIELD), _DEFAULT_SORT_ASC
 
 
 # --- Helper for safe ORM -> Pydantic conversion (Test/Mock compatible) ---
@@ -874,15 +878,17 @@ async def search_threads(
         stmt = stmt.where(ThreadORM.status == request.status)
 
     if request.metadata:
-        for key, value in request.metadata.items():
-            # Compare as JSONB so Python bools/ints/None match their JSON
-            # counterparts (str(True) == "True" was the previous bug).
-            stmt = stmt.where(ThreadORM.metadata_json[key] == cast(bindparam(None, value=value, type_=JSON), JSONB))
+        # JSONB containment: type-correct, deep-nested, GIN-indexable. Mirrors
+        # AssistantService.search_assistants for cross-endpoint consistency.
+        stmt = stmt.where(ThreadORM.metadata_json.op("@>")(request.metadata))
 
     offset = request.offset or 0
     limit = request.limit or 20
     column, asc = _resolve_sort(request)
-    stmt = stmt.order_by(column.asc() if asc else column.desc()).offset(offset).limit(limit)
+    direction = column.asc() if asc else column.desc()
+    # Secondary sort on thread_id keeps offset pagination stable when the
+    # primary sort key has duplicates (status buckets, microsecond ties).
+    stmt = stmt.order_by(direction, ThreadORM.thread_id.asc()).offset(offset).limit(limit)
 
     result = await session.scalars(stmt)
     rows = result.all()
--- a/libs/aegra-api/src/aegra_api/api/threads.py
+++ b/libs/aegra-api/src/aegra_api/api/threads.py
@@ -59,7 +59,12 @@ def _resolve_sort(request: ThreadSearchRequest) -> tuple[Any, bool]:
     field = _DEFAULT_SORT_FIELD
     asc = _DEFAULT_SORT_ASC
 
-    if request.order_by:
+    if request.sort_by:
+        candidate = request.sort_by.strip().lower()
+        if candidate in _ALLOWED_SORT_FIELDS:
+            field = candidate
+            asc = (request.sort_order or "desc").lower() == "asc"
+    elif request.order_by:
         parts = request.order_by.strip().split()
         if parts and parts[0].lower() in _ALLOWED_SORT_FIELDS:
             field = parts[0].lower()
--- a/libs/aegra-api/src/aegra_api/api/threads.py
+++ b/libs/aegra-api/src/aegra_api/api/threads.py
@@ -9,8 +9,10 @@ from uuid import uuid4
 
 import structlog
 from fastapi import APIRouter, Depends, HTTPException, Query
-from sqlalchemy import select
+from sqlalchemy import bindparam, cast, select
+from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.ext.asyncio import AsyncSession
+from sqlalchemy.types import JSON
 
 from aegra_api.core.active_runs import active_runs
 from aegra_api.core.auth_deps import auth_dependency, get_current_user
@@ -873,7 +875,12 @@ async def search_threads(
 
     if request.metadata:
         for key, value in request.metadata.items():
-            stmt = stmt.where(ThreadORM.metadata_json[key].as_string() == str(value))
+            # Compare as JSONB so Python bools/ints/None match their JSON
+            # counterparts (str(True) == "True" was the previous bug).
+            stmt = stmt.where(
+                ThreadORM.metadata_json[key]
+                == cast(bindparam(None, value=value, type_=JSON), JSONB)
+            )
 
     offset = request.offset or 0
     limit = request.limit or 20
--- a/libs/aegra-api/src/aegra_api/api/threads.py
+++ b/libs/aegra-api/src/aegra_api/api/threads.py
@@ -877,10 +877,7 @@ async def search_threads(
         for key, value in request.metadata.items():
             # Compare as JSONB so Python bools/ints/None match their JSON
             # counterparts (str(True) == "True" was the previous bug).
-            stmt = stmt.where(
-                ThreadORM.metadata_json[key]
-                == cast(bindparam(None, value=value, type_=JSON), JSONB)
-            )
+            stmt = stmt.where(ThreadORM.metadata_json[key] == cast(bindparam(None, value=value, type_=JSON), JSONB))
 
     offset = request.offset or 0
     limit = request.limit or 20
--- a/libs/aegra-api/src/aegra_api/api/threads.py
+++ b/libs/aegra-api/src/aegra_api/api/threads.py
@@ -42,6 +42,32 @@ logger = structlog.getLogger(__name__)
 thread_state_service = ThreadStateService()
 
 
+# --- Sort resolution for /threads/search ---
+
+_ALLOWED_SORT_FIELDS: frozenset[str] = frozenset({"created_at", "updated_at", "thread_id", "status"})
+_DEFAULT_SORT_FIELD = "created_at"
+_DEFAULT_SORT_ASC = False
+
+
+def _resolve_sort(request: ThreadSearchRequest) -> tuple[Any, bool]:
+    """Resolve (ORM column, is_ascending) for /threads/search.
+
+    Precedence: SDK-style ``sort_by`` (with optional ``sort_order``) wins over the
+    legacy ``order_by`` string. Unknown columns or malformed input silently fall
+    back to the default (``created_at DESC``) — never 500.
+    """
+    field = _DEFAULT_SORT_FIELD
+    asc = _DEFAULT_SORT_ASC
+
+    if request.order_by:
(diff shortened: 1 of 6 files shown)


libs/aegra-api/src/aegra_api/api/threads.py
```diff
@@ -42,6 +44,37 @@
...
+        if candidate in _ALLOWED_SORT_FIELDS:
+            field = candidate
+            asc = (request.sort_order or "desc").lower() == "asc"
+    elif request.order_by:
+        parts = request.order_by.strip().split()
+        if parts and parts[0].lower() in _ALLOWED_SORT_FIELDS:
+            field = parts[0].lower()
+            asc = len(parts) > 1 and parts[1].lower() == "asc"
+
+    return getattr(ThreadORM, field), asc
```
ibbybuilds: **precedence bug.** if `sort_by` is truthy but not in `_ALLOWED_SORT_FIELDS` (e.g. `sort_by="foo"`), the `elif` is skipped entirely and a valid `order_by="updated_at ASC"` in the same request is silently ignored. fix: only return early when `sort_by` actually matched, otherwise fall through to `order_by`.

```python
def _resolve_sort(request: ThreadSearchRequest) -> tuple[Any, bool]:
    if request.sort_by:
        candidate = request.sort_by.strip().lower()
        if candidate in _ALLOWED_SORT_FIELDS:
            asc = (request.sort_order or "desc").lower() == "asc"
            return getattr(ThreadORM, candidate), asc
    if request.order_by:
        parts = request.order_by.strip().split()
        if parts and parts[0].lower() in _ALLOWED_SORT_FIELDS:
            asc = len(parts) > 1 and parts[1].lower() == "asc"
            return getattr(ThreadORM, parts[0].lower()), asc
    return getattr(ThreadORM, _DEFAULT_SORT_FIELD), _DEFAULT_SORT_ASC
```

worth a test: `sort_by="foo"` + `order_by="updated_at ASC"` should sort by updated_at asc, currently sorts created_at desc.

libs/aegra-api/src/aegra_api/api/threads.py
```diff
@@ -842,11 +875,14 @@ async def search_threads(
 
     if request.metadata:
         for key, value in request.metadata.items():
-            stmt = stmt.where(ThreadORM.metadata_json[key].as_string() == str(value))
+            # Compare as JSONB so Python bools/ints/None match their JSON
+            # counterparts (str(True) == "True" was the previous bug).
+            stmt = stmt.where(ThreadORM.metadata_json[key] == cast(bindparam(None, value=value, type_=JSON), JSONB))
```
ibbybuilds: **swap to JSONB containment.** `assistant_service.search_assistants:250` already does `AssistantORM.metadata_dict.op("@>")(request.metadata)`. one clause instead of N, GIN-indexable (see the migration ask in the review body), and matches nested-value semantics correctly. per-key `==` requires whole-subtree equality at that key, `@>` does deep containment, which is what callers expect for filters like `metadata={"user": {"role": "admin"}}`. drops the `bindparam(None, ...)` ceremony entirely:

```python
if request.metadata:
    stmt = stmt.where(ThreadORM.metadata_json.op("@>")(cast(request.metadata, JSONB)))
```

also keeps thread search and assistant search behaviorally consistent. right now they diverge on nested filters.

libs/aegra-api/src/aegra_api/api/threads.py
```diff
@@ -842,11 +875,14 @@ async def search_threads(
...
 
     offset = request.offset or 0
     limit = request.limit or 20
-    stmt = stmt.order_by(ThreadORM.created_at.desc()).offset(offset).limit(limit)
+    column, asc = _resolve_sort(request)
+    stmt = stmt.order_by(column.asc() if asc else column.desc()).offset(offset).limit(limit)
```
ibbybuilds: **no stable tie-break.** sorting by `status` produces 4 distinct values across a user's threads (idle/busy/interrupted/error). with `offset/limit` pagination the row order within a status bucket is unspecified, so pages overlap or drop rows. `created_at` can also tie at microsecond resolution under bulk insert.

```python
stmt = stmt.order_by(column.asc() if asc else column.desc(), ThreadORM.thread_id.asc()).offset(offset).limit(limit)
```

worth a regression test: paginate `sort_by="status"` with `limit=2` across a seeded set of 6 threads and assert the union equals the full set with no overlap.

libs/aegra-api/src/aegra_api/models/threads.py
```diff
@@ -71,7 +71,18 @@ class ThreadSearchRequest(BaseModel):
     status: str | None = Field(None, description="Thread status filter (idle, busy, interrupted, error)")
     limit: int | None = Field(20, le=100, ge=1, description="Maximum results")
     offset: int | None = Field(0, ge=0, description="Results offset")
-    order_by: str | None = Field("created_at DESC", description="Sort order")
+    order_by: str | None = Field(
+        "created_at DESC",
+        description="Sort order (legacy single-field form, e.g. 'updated_at ASC')",
+    )
+    sort_by: str | None = Field(
```
ibbybuilds: **tighten `sort_by` to a Literal.** matches the strictness of `sort_order` directly below, gives clients a clean 422 on typos instead of silently sorting by `created_at DESC`, and lets you drop `_ALLOWED_SORT_FIELDS` entirely (Pydantic does the work).

```python
sort_by: Literal["thread_id", "status", "created_at", "updated_at"] | None = Field(
    None,
    description="Field to sort by (SDK-compatible). Takes precedence over order_by.",
)
```

update tests: `test_sdk_state_updated_at_falls_back` becomes `test_sdk_unknown_sort_by_returns_422`. the SQL-injection-style `sort_by` test moves to the 422 path; keep the matching test for `order_by` since that field stays permissive on purpose.

libs/aegra-api/src/aegra_api/models/threads.py
```diff
@@ -71,7 +71,18 @@ class ThreadSearchRequest(BaseModel):
     status: str | None = Field(None, description="Thread status filter (idle, busy, interrupted, error)")
     limit: int | None = Field(20, le=100, ge=1, description="Maximum results")
     offset: int | None = Field(0, ge=0, description="Results offset")
-    order_by: str | None = Field("created_at DESC", description="Sort order")
+    order_by: str | None = Field(
```
ibbybuilds, line 74: nit: add `deprecated=True` to the `order_by` Field so OpenAPI flags it for SDK clients. the field stays for backward compat but no new callers should adopt it.

```python
order_by: str | None = Field(
    "created_at DESC",
    deprecated=True,
    description="DEPRECATED: use sort_by + sort_order. Legacy single-field form, e.g. 'updated_at ASC'.",
)
```

ibbybuilds (changes requested):
hey @arturbagramyan1, thanks for digging into #302 and the three-commits separation, really clean. read through the full diff a few times. agree on all three bugs, tests are excellent.

before merging i want to push on a few things so this lands as the canonical search implementation. nothing huge, but enough that i'd like another round. specifics are in the inline comments.

## must-fix (see inline)

1. precedence bug in `_resolve_sort` — invalid `sort_by` silently swallows valid `order_by`.
2. swap metadata predicate to JSONB `@>` containment, matching `assistant_service`.
3. tighten `sort_by` to a `Literal`, drop the silent fallback branch.
4. add a stable secondary tie-break on `thread_id` for pagination correctness.
5. add a GIN `jsonb_path_ops` index migration to back the new `@>` predicate.

plus a nit: mark `order_by` `deprecated=True` in the Field so OpenAPI flags it.

## keep as-is

- `order_by` legacy field stays (back-compat).
- silent fallback on malformed `order_by` is the right call (legacy permissive parsing).
- the JSONB-cast bool fix at the e2e level is great.
- commit separation, test coverage, type annotations, all clean.

arturbagramyan1:
thanks for the detailed review @ibbybuilds — agreed on all five must-fix items and the deprecated nit. pushed two commits: \`df960ab\` (code + tests) and \`6bff89f\` (migration). breakdown by item:

**1. precedence in \`_resolve_sort\`** — fixed. structure now early-returns only on actual match, so an invalid \`sort_by\` cannot swallow a valid \`order_by\`. side note: with item 4 in place the bad combo is unreachable from the request boundary anyway, since Pydantic 422s before the helper runs. left the early-return shape regardless because it's the right structure for the function.

**2. swap to \`@>\` containment** — done. now \`stmt.where(ThreadORM.metadata_json.op(\"@>\")(request.metadata))\`, matching \`AssistantService.search_assistants:250\`. nested-value filters now work as expected, dropped the \`bindparam(None, ..., type_=JSON)\` ceremony entirely.

**3. stable tie-break on \`thread_id\`** — done. \`order_by(direction, ThreadORM.thread_id.asc())\`. added an e2e test \`test_search_pagination_stable_with_tied_sort_key_e2e\` that seeds 6 threads with the same status, paginates with \`limit=2\` × 3 offsets, and asserts the union is a strict partition of the seed (no overlap, no drops, no dupes).

**4. tighten \`sort_by\` to Literal** — done. \`Literal[\"thread_id\", \"status\", \"created_at\", \"updated_at\"]\`. typos and \`state_updated_at\` now 422 with the standard Pydantic field error mentioning the allowed values. tests updated:
- \`test_sdk_state_updated_at_falls_back_silently\` and \`test_sdk_unknown_sort_by_falls_back\` (both unit) → removed; can't reach the helper.
- \`test_search_sdk_state_updated_at_falls_back\` (integration) → \`test_search_sdk_state_updated_at_returns_422\`.
- added \`test_search_invalid_sort_by_returns_422\` covering the precedence regression at the request boundary.

**5. GIN \`jsonb_path_ops\` migration** — done. new migration \`20260428000000_add_gin_index_on_thread_metadata.py\` (revision \`b7c8d9e0f123\`, chains off \`a1b2c3d4e5f6\`). verified locally:

\`\`\`
docker exec aegra-postgres-1 psql -U user -d aegra -c \"\\d+ thread\"
...
\"idx_thread_metadata_gin\" gin (metadata_json jsonb_path_ops)
\`\`\`

**nit. \`deprecated=True\` on \`order_by\`** — done. one detail worth flagging: Pydantic v2's \`deprecated=True\` not only marks the field in JSON schema (what we want for OpenAPI) but also fires a runtime \`DeprecationWarning\` on every attribute access. since the handler still reads \`order_by\` to honour the back-compat path, that produced a per-request warning. wrapped just the handler-side read in \`warnings.catch_warnings()\` so the deprecation reaches SDK consumers via the schema flag, not server logs.

quick verification against the live container:

\`\`\`
$ curl -s -X POST localhost:2026/threads/search -H 'content-type: application/json' \
    -d '{\"sort_by\":\"definitely_not_a_column\",\"order_by\":\"updated_at ASC\"}' | jq .detail
[{
  \"type\": \"literal_error\",
  \"loc\": [\"body\", \"sort_by\"],
  \"msg\": \"Input should be 'thread_id', 'status', 'created_at' or 'updated_at'\",
  ...
}]

$ uv run --package aegra-api pytest libs/aegra-api/tests/unit libs/aegra-api/tests/integration
============================ 1184 passed in 13.06s ============================

$ pytest libs/aegra-api/tests/e2e/test_threads/test_search.py
============================== 7 passed in 9.35s ==============================
\`\`\`

ready for another look.

ibbybuilds (approved):

PR #323, opened 2026-04-23
fix(api): SSE keepalive and transport-layer disconnect handling

Witaly3:
## Description

  Production agents with long silent graph nodes (e.g. holding an upstream WebSocket for ~60s) were being
  interrupted by Aegra. The SSE response had no keepalive, so idle proxies dropped the connection; Starlette turned
  that into a `CancelledError` inside the streaming generator, and the generator's `cancel_on_disconnect` branch
  fired `broker_manager.request_cancel` — misreading a proxy idle-drop as a real client disconnect.

  This PR fixes it by migrating SSE to `sse_starlette.EventSourceResponse` with a `: heartbeat` ping (byte-for-byte
  identical to `langgraph-api==0.8.0`) and moving cancel-on-disconnect from the generator to the transport layer, so
   only a real ASGI `http.disconnect` cancels the run.

  ## Type of Change

  - [ ] `feat`: New feature
  - [x] `fix`: Bug fix
  - [ ] `ci`: CI/CD changes

  ## Related Issues

  

  ## Changes Made

  - Migrated 3 SSE endpoints (`POST /threads/{id}/runs/stream`, `GET /threads/{id}/runs/{rid}/stream`, `POST
  /runs/stream`) from `StreamingResponse` to `sse_starlette.EventSourceResponse`. `/wait` and `/join` keep their
  `\n`-byte JSON heartbeat — different content-type contract.
  - Emit `: heartbeat` comment frames via `ping_message_factory`, byte-for-byte identical to `langgraph-api==0.8.0`.
   Interval configurable via `KEEPALIVE_INTERVAL_SECS` (default 5s). Added `sse_ping_interval_secs` computed field
  with `max(1, int(...))` clamping (sse-starlette requires int).
  - Removed `cancel_on_disconnect` from `streaming_service.stream_run_execution`; the generator's `CancelledError`
  handler is now side-effect-free. Cancellation is wired through
  `EventSourceResponse.client_close_handler_callable`, firing only on real ASGI `http.disconnect`.
  - The reconnect endpoint `GET /threads/{id}/runs/{rid}/stream` deliberately omits `client_close_handler_callable`
  — a single client disconnecting must not cancel the shared run, since other consumers (`/join` or another
  `/stream`) may still be attached.
  - `stateless_stream_run` steals both `body_iterator` and `client_close_handler_callable` from the inner response
  and rewraps them in an outer `EventSourceResponse`, so ping and disconnect run against the cleaned-up iterator
  while disconnect-cancel still reaches the inner broker.
  - Added `sse-starlette>=3.3.4,<4.0.0` dependency.
  - Updated `docs/guides/streaming.mdx` with the new keepalive/disconnect behavior.

  ## Testing

  - [ ] E2E tests added/updated

  New tests:

  - `tests/integration/test_sse_keepalive.py` (3 tests):
    - `test_keepalive_pings_during_silent_broker` — counts `: heartbeat` on the wire during a silent broker.
    - `test_client_disconnect_triggers_close_handler` — drives ASGI directly (httpx.ASGITransport does not
  synthesize `http.disconnect`) and asserts the handler fires exactly once.
    - `test_cancelled_error_no_longer_requests_cancel` — regression for the original bug; monkeypatches
  `request_cancel` and asserts it is never called on `CancelledError`.
  - `tests/unit/test_api/test_runs_streaming.py` (+2 parametrized) — disconnect-handler wiring across
  `on_disconnect` values and broker-error swallowing.
  - `tests/unit/test_api/test_stateless_runs.py` — mocks updated to `EventSourceResponse` + bytes iterators.
  - `tests/unit/test_settings.py::TestSsePingIntervalSecs` — default, truncation, sub-second clamping.

  Full suite: **1174 passed**, lint + type-check clean.

  Manual verification against `docker compose up -d`:

  | Endpoint | `: heartbeat` hits | legacy `: ping` hits |
  |---|---|---|
  | `POST /threads/{id}/runs/stream` | ≥1 | 0 |
  | `GET /threads/{id}/runs/{rid}/stream` | ≥2 | 0 |
  | `POST /runs/stream` (stateless) | ≥1 | 0 |

  Raw wire frame observed: `: heartbeat\r\n\r\n` (15 bytes) — matches `langgraph-api==0.8.0` byte-for-byte.

  ## Checklist

  N/A — server-side SSE wire change.

  ## Additional Notes

  **Client compatibility (important for upstream adopters).** W3C EventSource spec §9.2.6 requires clients to ignore
   lines starting with `:`. Verified in the relevant SSE parsers used against Aegra:

  - `langgraph_sdk/sse.py` — filters `:` lines
  - `httpx_sse/_decoders.py` — filters `:` lines
  - Browser `EventSource` — spec-compliant in all major engines

  Only self-written parsers without a `:` filter would need updating — one such parser exists in
  `tests/e2e/test_streaming/test_streaming_error_e2e.py` but currently silently skips heartbeats by accident (safe).

  **Not migrated.** `/runs/{id}/join` and `/runs/wait` (both threaded and stateless) still use `StreamingResponse`
  with a `\n`-byte heartbeat in `run_waiters.py` — that's correct for their `application/json` contract.

  **DEBUG logs from `sse_starlette.sse._ping` are deliberately left in** — useful in debug mode, and not overridable
   without forking `_ping`.

## Summary by CodeRabbit

* **New Features**
  * SSE endpoints now emit periodic keepalive heartbeats to reduce idle-proxy disconnects.
  * Keepalive interval exposed as an integer ping setting (truncated/clamped to >=1s).

* **Behavior**
  * Client disconnects cancel runs only for endpoints configured to do so; reconnect-style streams won’t cancel shared runs on single-client disconnects.
  * Errors in streams are emitted as structured SSE error events; heartbeat lines are SSE comments and ignored by standard clients.

* **Documentation**
  * Guide updated with keepalive and cancellation interoperability details.

* **Tests**
  * New integration, unit and e2e tests validating keepalives, disconnect handling, and cancellation semantics.

<h3>Greptile Summary</h3>

- Migrates three SSE endpoints from `StreamingResponse` to `sse_starlette.EventSourceResponse` with `: heartbeat` keepalive pings (default 5 s, configurable via `KEEPALIVE_INTERVAL_SECS`), fixing silent-agent interruption caused by idle-proxy timeouts firing a spurious `CancelledError` that was being misread as a real client disconnect.
- Moves `cancel_on_disconnect` logic from the streaming generator to `client_close_handler_callable` on the transport layer, so only a genuine ASGI `http.disconnect` event cancels the run; proxy drops no longer trigger `broker_manager.request_cancel`. The reconnect endpoint (`GET .../stream`) deliberately omits the close handler.
- Fixes a latent `UnboundLocalError` in `stateless_wait_for_run`'s `_wrapped_iterator` (missing `completed = False` initializer) and narrows broad `except Exception` cleanup catches to a typed `_CLEANUP_ERRORS` tuple so programmer errors surface instead of being silently swallowed.

Safe to merge — the core transport-layer disconnect fix is architecturally sound, well-tested, and the latent UnboundLocalError in the wait endpoint is also resolved.

No P0 or P1 findings. The design cleanly separates SSE keepalive from run cancellation, all three path variants (cancel/continue/reconnect) are covered by unit and integration tests, and the previous review thread findings have been addressed in this revision. 1174 tests pass.

| Filename | Overview |
|----------|----------|
| libs/aegra-api/src/aegra_api/core/sse.py | New heartbeat_factory, sse_to_bytes, and make_sse_response helpers that centralize SSE defaults; sse_to_bytes wraps the inner generator with aclosing for deterministic cleanup |
| libs/aegra-api/src/aegra_api/api/runs.py | Migrated create_and_stream_run and stream_run from StreamingResponse to EventSourceResponse; disconnect-cancel moved to _cancel_on_client_close closure wired via client_close_handler_callable; reconnect endpoint correctly omits the close handler |
| libs/aegra-api/src/aegra_api/api/stateless_runs.py | Narrow _CLEANUP_ERRORS tuple replaces bare except Exception; new _wrapped_iterator correctly handles completed/slow-client-abort/early-disconnect paths; Content-Location header is correctly extracted for run_id lookup |
| libs/aegra-api/src/aegra_api/services/streaming_service.py | Removed cancel_on_disconnect parameter; CancelledError handler is now side-effect-free; error event now emits structured {error: TypeName, message: execution failed} to avoid leaking internal details |
| libs/aegra-api/src/aegra_api/settings.py | Added sse_ping_interval_secs computed field that clamps the float KEEPALIVE_INTERVAL_SECS to an int >= 1, matching sse-starlette's requirement |
| libs/aegra-api/tests/integration/test_sse_keepalive.py | Three focused integration tests: heartbeat counting during silent broker, ASGI-driven http.disconnect firing close handler, and regression for CancelledError no longer calling request_cancel |
| libs/aegra-api/tests/unit/test_api/test_runs_streaming.py | New parametrized tests cover disconnect-handler wiring for all on_disconnect values and broker-error swallowing; stream_run confirmed to never wire a close handler |
| libs/aegra-api/tests/unit/test_api/test_stateless_runs.py | Updated mocks to EventSourceResponse + bytes iterators; new tests for early disconnect, slow-client abort (finished/active run), missing header, and programmer-error propagation |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Client
    participant ASGI as ASGI / sse-starlette
    participant Handler as Endpoint Handler
    participant SSE as sse_to_bytes + stream_run_execution
    participant Broker as RunBroker

    Client->>ASGI: POST /threads/{id}/runs/stream
    ASGI->>Handler: create_and_stream_run()
    Handler->>Handler: build _cancel_on_client_close (if cancel_on_disconnect)
    Handler->>ASGI: EventSourceResponse(body=sse_to_bytes(...), ping=5s, close_handler=...)
    ASGI-->>Client: HTTP 200 text/event-stream

    loop Every 5 s (idle)
        ASGI-->>Client: : heartbeat CRLF CRLF
    end

    Broker-->>SSE: put(event_data)
    SSE-->>ASGI: bytes (pre-formatted SSE)
    ASGI-->>Client: event chunk

    alt Real http.disconnect (client closes)
        Client->>ASGI: http.disconnect
        ASGI->>Handler: client_close_handler_callable(msg)
        Handler->>Broker: broker_manager.request_cancel(run_id)
    else Proxy idle-drop (no real disconnect)
        Note over ASGI,Handler: CancelledError in generator -> re-raise only, no request_cancel
    end
```

<sub>Reviews (2): Last reviewed commit: ["fix(api): address SSE keepalive follow-u..."](https://github.com/aegra/aegra/commit/207359ec7c452f95e063e0dd80ad7cb4c65b081b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30818441)</sub>

### DIFF
--- a/docs/guides/streaming.mdx
+++ b/docs/guides/streaming.mdx
@@ -183,6 +183,8 @@ async for chunk in client.runs.stream(
 | `"cancel"` (default) | Cancel the run when client disconnects |
 | `"continue"` | Run continues in the background; reconnect later to get results |
 
+The server sends SSE keepalive comments (`: heartbeat`) every `KEEPALIVE_INTERVAL_SECS` (default 5s), so idle proxies (Nginx 60s, AWS ALB, Cloudflare) won't drop long-running silent agents — for example, a graph node holding an upstream WebSocket without emitting events. The wire-format matches LangGraph Platform, so existing SSE clients (the LangGraph SDK, browser `EventSource`, `httpx-sse`) silently ignore these lines per the W3C spec. Only a real client disconnect — not an idle proxy timeout — triggers automatic cancellation.
+
 ## Background runs
 
 For long-running tasks, you can create a run in the background and check on it later:
--- a/docs/guides/streaming.mdx
+++ b/docs/guides/streaming.mdx
@@ -183,7 +183,7 @@ async for chunk in client.runs.stream(
 | `"cancel"` (default) | Cancel the run when client disconnects |
 | `"continue"` | Run continues in the background; reconnect later to get results |
 
-The server sends SSE keepalive comments (`: heartbeat`) every `KEEPALIVE_INTERVAL_SECS` (default 5s), so idle proxies (Nginx 60s, AWS ALB, Cloudflare) won't drop long-running silent agents — for example, a graph node holding an upstream WebSocket without emitting events. The wire-format matches LangGraph Platform, so existing SSE clients (the LangGraph SDK, browser `EventSource`, `httpx-sse`) silently ignore these lines per the W3C spec. Only a real client disconnect — not an idle proxy timeout — triggers automatic cancellation.
+The server sends SSE keepalive comments (`: heartbeat`) every `KEEPALIVE_INTERVAL_SECS` (default 5s, truncated to whole seconds, minimum 1s), so idle proxies (Nginx 60s, AWS ALB, Cloudflare) won't drop long-running silent agents — for example, a graph node holding an upstream WebSocket without emitting events. The wire-format matches LangGraph Platform, so existing SSE clients (the LangGraph SDK, browser `EventSource`, `httpx-sse`) silently ignore these lines per the W3C spec. Only a real client disconnect — not an idle proxy timeout — triggers automatic cancellation.
 
 ## Background runs
 
--- a/libs/aegra-api/pyproject.toml
+++ b/libs/aegra-api/pyproject.toml
@@ -22,6 +22,7 @@ dependencies = [
     "uvicorn>=0.35.0",
     "pydantic>=2.11.7",
     "pydantic-settings>=2.12.0",
+    "sse-starlette>=3.3.4,<4.0.0",
 
     # LangGraph
     "langgraph>=1.0.3",
--- a/libs/aegra-api/src/aegra_api/api/runs.py
+++ b/libs/aegra-api/src/aegra_api/api/runs.py
@@ -2,23 +2,26 @@
 
 import asyncio
 import contextlib
-from collections.abc import AsyncIterator
+from collections.abc import AsyncIterator, MutableMapping
 from datetime import UTC, datetime
+from typing import Any
 
 import structlog
 from fastapi import APIRouter, Depends, Header, HTTPException, Query
 from fastapi.responses import StreamingResponse
 from sqlalchemy import delete, select, update
 from sqlalchemy.ext.asyncio import AsyncSession
+from sse_starlette import EventSourceResponse
 
 from aegra_api.core.active_runs import active_runs
 from aegra_api.core.auth_deps import auth_dependency, get_current_user
 from aegra_api.core.auth_handlers import build_auth_context, handle_event
 from aegra_api.core.orm import Run as RunORM
 from aegra_api.core.orm import _get_session_maker, get_session
-from aegra_api.core.sse import create_end_event, get_sse_headers
+from aegra_api.core.sse import create_end_event, get_sse_headers, heartbeat_factory, sse_to_bytes
 from aegra_api.models import Run, RunCreate, RunStatus, User
 from aegra_api.models.errors import CONFLICT, NOT_FOUND, SSE_RESPONSE
+from aegra_api.services.broker import broker_manager
 from aegra_api.services.run_preparation import _prepare_run
 from aegra_api.services.run_waiters import TERMINAL_STATES, encode_output, heartbeat_wait_body
 from aegra_api.services.streaming_service import streaming_service
@@ -81,7 +84,7 @@ async def create_and_stream_run(
     request: RunCreate,
     user: User = Depends(get_current_user),
     session: AsyncSession = Depends(get_session),
-) -> StreamingResponse:
+) -> EventSourceResponse:
     """Create a new run and stream its execution via SSE.
 
     Returns a `text/event-stream` response with Server-Sent Events. Each
@@ -91,6 +94,10 @@ async def create_and_stream_run(
     Set `on_disconnect` to `"continue"` if the run should keep executing
     after the client disconnects (default is `"cancel"`). Use `stream_mode`
     to control which event types are emitted.
+
+    A periodic SSE keepalive comment is sent every
+    ``KEEPALIVE_INTERVAL_SECS`` so idle proxies don't drop long-running
+    silent nodes (e.g. agents holding an upstream WebSocket).
     """
     run_id, run, _job = await _prepare_run(session, thread_id, request, user, initial_status="pending")
 
@@ -99,13 +106,23 @@ async def create_and_stream_run(
     # set on_disconnect="continue" if they want the task to continue.
     cancel_on_disconnect = (request.on_disconnect or "cancel").lower() == "cancel"
 
-    return StreamingResponse(
-        streaming_service.stream_run_execution(
-            run,
-            None,
-            cancel_on_disconnect=cancel_on_disconnect,
-        ),
-        media_type="text/event-stream",
+    async def _cancel_on_client_close(_msg: MutableMapping[str, Any]) -> None:
+        try:
+            await broker_manager.request_cancel(run_id, "cancel")
+        except Exception:
+            # Swallow to avoid cascading the failure through sse-starlette's
+            # task group. If the broker is unreachable we can't stop the run;
+            # it will either complete normally (result persists to Postgres)
+            # or be reaped when its worker lease expires.
+            logger.exception("Failed to cancel run on client disconnect", run_id=run_id)
+
+    close_handler = _cancel_on_client_close if cancel_on_disconnect else None
+
+    return EventSourceResponse(
+        sse_to_bytes(streaming_service.stream_run_execution(run, None)),
+        ping=settings.app.sse_ping_interval_secs,
+        ping_message_factory=heartbeat_factory,
+        client_close_handler_callable=close_handler,
         headers={
             **get_sse_headers(),
             "Location": f"/threads/{thread_id}/runs/{run_id}/stream",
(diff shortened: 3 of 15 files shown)


ibbybuilds (changes requested):
took a closer look at this. overall it lands the fix in the right place and the regression tests are solid. few things i want cleaned up before i merge.

## must-fix

**1. bare `except Exception:` in 6 spots**

runs.py:111 and stateless_runs.py:75, 100, 134, 161, 233, 281. CLAUDE.md is pretty clear about not doing this. for the broker/cancel paths narrow it to `(RedisError, ConnectionError, TimeoutError, OSError)`. don't want random programmer errors getting swallowed forever.

**2. slow-client path leaks ephemeral threads**

the `completed = True` flag only flips when the `async for` exits naturally. if the body iterator gets `aclose()`'d mid-stream because of a send-timeout (slow consumer, dead proxy in front), `completed` stays False and the thread never gets deleted. the new disconnect tests don't cover this.

i think the cleanest fix is in `_wrapped_iterator`'s finally. when `completed=False` but the run actually finished (check broker state), schedule the delayed cleanup task. that way early-disconnect still keeps the thread for resume, and slow-client-abort still cleans up.

**3. `str(e)` going to the wire in streaming_service.py:119**

```python
except Exception as e:
    yield create_error_event(str(e))
```

`str(e)` can leak file paths, internal IDs, frame hints. `create_error_event` already supports the structured form, so:

```python
yield create_error_event({"error": type(e).__name__, "message": "execution failed"})
logger.exception("stream execution failed", run_id=run_id)
```

also add a quick comment about why we swallow here instead of re-raising. CLAUDE.md says log or re-raise.

## should-fix (not blocking)

**4. duplicate `EventSourceResponse(...)` builder**

runs.py and stateless_runs.py construct the response with identical kwargs (ping, factory, headers). worth extracting to `core/sse.py`:

```python
def make_sse_response(body, *, headers, close_handler=None) -> EventSourceResponse: ...
```

**5. wrap `stream_run_execution(...)` in `contextlib.aclosing`**

deterministic close on cancel. ensures broker cleanup hooks fire even if the consumer gets killed before the generator's finally runs.

**6. missing prod-mode E2E**

dev-mode coverage is good. would like a `@pytest.mark.prod_only` test for disconnect into cross-instance cancel via Redis, since the cancel path now goes through `broker_manager.request_cancel`. catches regressions in pub/sub fan-out.

## verified clean

settings live-read at request time (not frozen at import), confirmed via test_settings.py float to int clamp. type annotations, imports at top, no mutable defaults all clean. auth/tenancy untouched. `\n` vs `\r\n` SSE separator both spec-compliant. `KEEPALIVE_INTERVAL_SECS` configurable plus validated is a nice touch.

## blast radius

every SSE endpoint. no effect on non-streaming runs, crons, webhooks, worker heartbeat. roughly 3KB/s outbound at 1k concurrent streams with default 5s ping. negligible.

requesting changes on the 3 must-fix. 4 to 6 can be follow-ups if you want to keep this PR focused.

ibbybuilds (approved):
took another pass after the rework. all 3 must-fix from my last review are landed and the slow-client cleanup branch is a thoughtful addition i wasn't expecting. nice work.

## verified addressed

**1. broad except** -> narrowed to `_CLEANUP_ERRORS = (RedisError, SQLAlchemyError, OSError)` in stateless_runs and `(RedisError, OSError)` in the disconnect handler. tuple is centralized so future additions are one-line.

**2. slow-client thread leak** -> `_run_finished` + `_schedule_thread_cleanup` covers the abort-after-run-finished case. acknowledged the prod-mode cross-instance gap with a clear TODO for an orphan-thread sweeper. fine as a known limitation, will track separately.

**3. str(e) on the wire** -> fixed in streaming_service.py:117. emits `{error: TypeName, message: "execution failed"}`, logs full exception server-side. clean.

**4. shared EventSourceResponse builder** -> `make_sse_response` in core/sse.py with shared ping/factory defaults.

**5. aclosing wrap** -> `sse_to_bytes` wraps in `contextlib.aclosing`. deterministic cleanup confirmed.

## remaining nits, non-blocking

- coderabbit's open thread asks for an explicit reconnect-path regression test asserting `stream_run` never wires a close handler. covered indirectly via the integration test on the create endpoint. one line in the unit suite would close that thread.
- codeql 96 + 97 on test file lines 191/243 are false positives. await on a task that was just cancel()'d IS the effect (cleanup of the suppressed CancelledError). dismiss with reason.
- greptile P1 on the disconnect handler scope: their theoretical RuntimeError-in-task-group concern is real but `request_cancel` already handles RedisError internally and falls back to a local task.cancel + broker put. none of which naturally raise RuntimeError. narrow over broad is the right call.
- prod-mode E2E for cross-instance disconnect-cancel via Redis pub/sub still missing. acceptable as follow-up.

## blast radius reconfirmed

every SSE endpoint, plus the inner EventSourceResponse -> outer rewrap trick in stateless_stream_run. iterator + close-handler stealing is unusual but well-commented. tests cover the cancel/continue/reconnect split.

LGTM.

PR #325, opened 2026-04-25
fix(store): ignore empty namespace query values

mohankumarelec:
## Summary
- Ignore empty `namespace` query values in `GET /store/items` so `namespace=` is treated as no namespace.
- Add an integration regression test for `GET /store/items?namespace=&key=...`.

## Test plan
- [x] `uv run --package aegra-api pytest libs/aegra-api/tests/integration/test_api/test_store_crud.py`

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **Bug Fixes**
  * Requests with empty or list-style namespace parameters now behave like requests without a namespace, ensuring consistent scoping.

* **Tests**
  * Added integration tests verifying empty namespace parameters trigger the expected scoping and store operations.

* **Refactor**
  * Consolidated namespace normalization to ensure consistent behavior across store operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/aegra/aegra/pull/325)

<h3>Greptile Summary</h3>

This PR fixes a bug where an empty namespace query parameter was not treated as "no namespace," causing incorrect store scoping. It extracts a shared `_normalize_namespace` helper and applies it consistently across the GET and DELETE endpoints, with regression tests for both.

- **`store.py`**: New `_normalize_namespace(value)` handles `str | list[str] | None` → `list[str]`, filtering empty parts; applied to `get_store_item` and both code paths of `delete_store_item`.
- **`test_store_crud.py`**: Two new integration tests assert that an empty namespace query param results in `apply_user_namespace_scoping` receiving `[]`, and the store backend is called with the correct user-scoped namespace tuple.

Safe to merge — the change is a targeted normalization fix with no behavioral regressions on existing paths.

The refactoring is minimal and well-scoped: a single helper function replaces inline logic that was already correct for the non-empty case, and the new tests directly cover the previously broken empty-string path. All three affected call sites (GET query param, DELETE query param, DELETE body) now go through the same normalization, and `put_store_item` is intentionally unchanged since it operates on an already-validated JSON body.

| Filename | Overview |
|----------|----------|
| libs/aegra-api/src/aegra_api/api/store.py | Extracts `_normalize_namespace` helper and applies it to `get_store_item` (query param) and `delete_store_item` (both query param and body paths), fixing the empty-string namespace bug. |
| libs/aegra-api/tests/integration/test_api/test_store_crud.py | Adds two integration regression tests confirming that `?namespace=` is treated as no namespace for GET and DELETE, using `patch(..., wraps=...)` spies to assert the scoping function receives an empty list. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A[Incoming Request] --> B{Has body?}
    B -- Yes --> C[body.namespace]
    B -- No --> D[namespace query param]
    C --> E[_normalize_namespace]
    D --> E
    E --> F{Type?}
    F -- str --> G["split('.'), filter empty"]
    F -- list --> H["filter empty strings"]
    F -- None --> I["return []"]
    G --> J[list of str]
    H --> J
    I --> J
    J --> K[apply_user_namespace_scoping]
    K --> L["['users', user_id, ...namespace]"]
    L --> M[Store Operation]
```

<sub>Reviews (2): Last reviewed commit: ["fix(store): normalize namespace in delet..."](https://github.com/aegra/aegra/commit/057d760ffc4cc27e3adda3f5fc41a3e63a7220c1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30666164)</sub>

### DIFF
--- a/libs/aegra-api/src/aegra_api/api/store.py
+++ b/libs/aegra-api/src/aegra_api/api/store.py
@@ -82,7 +82,7 @@ async def get_store_item(
     if isinstance(namespace, str):
         ns_list = [part for part in namespace.split(".") if part]
     elif isinstance(namespace, list):
-        ns_list = namespace
+        ns_list = [part for part in namespace if part]
     else:
         ns_list = []
 
--- a/libs/aegra-api/src/aegra_api/api/store.py
+++ b/libs/aegra-api/src/aegra_api/api/store.py
@@ -77,17 +77,8 @@ async def get_store_item(
         if "key" in filters:
             key = filters["key"]
 
-    # Accept SDK-style dotted namespaces or list
-    ns_list: list[str]
-    if isinstance(namespace, str):
-        ns_list = [part for part in namespace.split(".") if part]
-    elif isinstance(namespace, list):
-        ns_list = [part for part in namespace if part]
-    else:
-        ns_list = []
-
     # Apply user namespace scoping
-    scoped_namespace = apply_user_namespace_scoping(user.identity, ns_list)
+    scoped_namespace = apply_user_namespace_scoping(user.identity, _normalize_namespace(namespace))
 
     store = db_manager.get_store()
 
@@ -120,7 +111,7 @@ async def delete_store_item(
     else:
         if key is None:
             raise HTTPException(422, "Missing 'key' parameter")
-        ns = namespace or []
+        ns = _normalize_namespace(namespace)
         k = key
 
     # Authorization check
@@ -231,6 +222,15 @@ async def list_namespaces(
     )
 
     return StoreListNamespacesResponse(namespaces=[list(ns) for ns in result])
+
+
+def _normalize_namespace(value: str | list[str] | None) -> list[str]:
+    """Normalize namespace input to a clean list, filtering out empty parts."""
+    if isinstance(value, str):
+        return [part for part in value.split(".") if part]
+    if isinstance(value, list):
+        return [part for part in value if part]
+    return []
 
 
 def apply_user_namespace_scoping(user_id: str, namespace: list[str]) -> list[str]:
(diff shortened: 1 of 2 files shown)


mohankumarelec:
Hey @ibbybuilds , 
Awesome work on this project thanks a lot. Have been using it extensively and its amazing. Found a small bug when i tried to invoke the store endpoint from langgraph-sdk. Tested the fix and it works good. please feel free to edit if necessary.

ibbybuilds (changes requested):
small clean fix, lands in the right place. one thing i want addressed before merge plus a small test nit.

## must-fix

**delete endpoint has the same bug — store.py:123**

```python
ns = namespace or []
```

if a client sends `?namespace=&key=x`, fastapi parses `namespace=[""]` which is truthy, so the fallback never kicks in and `("",)` reaches `store.adelete`. same shape as the get bug this PR fixes.

```python
ns = [part for part in (namespace or []) if part]
```

worth doing in this PR so we don't ship a half-fix. ideally extract a tiny helper since we now have the same filtering logic in two places:

```python
def _normalize_namespace(value: str | list[str] | None) -> list[str]:
    if isinstance(value, str):
        return [part for part in value.split(".") if part]
    if isinstance(value, list):
        return [part for part in value if part]
    return []
```

## nit

**test only asserts on the post-scoping tuple**

`call_args[0][0] == ("users", "test-user")` would pass even if the empty-string filtering didn't happen, since `apply_user_namespace_scoping` happens to produce the same tuple either way. add a direct assert that the scoping function was called with `[]` (no empty string), so the test actually pins the behavior we just fixed.

other than that, looks good. fix delete + tighten the assert and i'll merge.

mohankumarelec:
@ibbybuilds , done

ibbybuilds:
@mohankumarelec please resolve greptile and other ai review comments, and fix linting issues.

mohankumarelec:
Done @ibbybuilds , Thanks

ibbybuilds (approved):
LGTM!

PR #327, opened 2026-04-25
[db] Add POSTGRES_SCHEMA support for schema-isolated deployments

mohankumarelec:
## Summary

Adds a new `POSTGRES_SCHEMA` env var that lets you run all Aegra tables inside a dedicated PostgreSQL schema instead of `public`. This is useful when you're sharing a database across multiple services or tenants and dont want everything lumped into the same schema.

When `POSTGRES_SCHEMA` is set:
- SQLAlchemy connections get `search_path` set via `server_settings`
- LangGraph pool connections get `search_path` via a per-connection `configure` callback
- The schema is auto-created (`CREATE SCHEMA IF NOT EXISTS`) on startup
- Alembic migrations create the schema and set `search_path` before running, and the version table is placed in the target schema
- The `uuid-ossp` extension is explicitly created in the `public` schema (extensions are cluster-wide, avoids issues when running in a non-public schema)

When not set (default), behavior is completely unchanged.

## Changes
- `settings.py` — new `POSTGRES_SCHEMA: str | None = None` field
- `database.py` — schema creation + search_path for both SQLAlchemy and LangGraph pools
- `alembic/env.py` — schema-aware offline and online migrations with `version_table_schema`
- `alembic/versions/initial_schema.py` — `uuid-ossp` extension pinned to public schema
- Tests — added 11 new unit tests covering settings defaults and all schema isolation paths in DatabaseManager

## Test plan
- [x] `POSTGRES_SCHEMA` defaults to `None`
- [x] `POSTGRES_SCHEMA` can be set via env var
- [x] SQLAlchemy `connect_args` includes `search_path` when schema is set
- [x] SQLAlchemy `connect_args` has no `server_settings` when schema is `None`
- [x] LangGraph pool gets `configure` callback when schema is set
- [x] LangGraph pool `configure` is `None` when schema is not set
- [x] `CREATE SCHEMA IF NOT EXISTS` runs on init
- [x] No schema creation when `POSTGRES_SCHEMA` is unset
- [x] Configure callback executes correct `SET search_path`
- [x] PgBouncer compatibility preserved alongside schema settings
- [x] All 912 unit tests pass, lint + format clean

## Summary by CodeRabbit

* **New Features**
  * Added PostgreSQL schema isolation support enabling multi-tenant database separation.
  * Database initialization now automatically manages schema creation and search path configuration.

* **Tests**
  * Added comprehensive test coverage for schema isolation functionality.

### DIFF
--- a/libs/aegra-api/alembic/env.py
+++ b/libs/aegra-api/alembic/env.py
@@ -4,7 +4,7 @@ import asyncio
 import threading
 from logging.config import fileConfig
 
-from sqlalchemy import pool
+from sqlalchemy import pool, text
 from sqlalchemy.engine import Connection
 from sqlalchemy.ext.asyncio import async_engine_from_config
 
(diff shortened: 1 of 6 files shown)


ibbybuilds (changes requested):
thanks @mohankumarelec, the mechanism here is mostly right. asyncpg `server_settings` is the correct way to set search_path on the SQLAlchemy pool, the psycopg `configure` callback is the right per-connection approach for the LangGraph pool, the schema name goes through `Identifier` so no injection, and making alembic schema-aware with `version_table_schema` is coherent.

blockers:

1. `POSTGRES_SCHEMA` isn't in either `.env.example` or the cli template, and it's not in `docs/guides/deployment.mdx`. any new env var has to land in all of those, it's a hard project rule.
2. `check-commits` is failing and the branch is conflicting. this is stale (the CI runs are from a while back), so it needs a full rebase against current main, which has since gained the multi-host HA / two-URL DB handling.

should-fix:
- the two-URL interaction isn't addressed. `server_settings` on asyncpg is fine, but the migration precheck path runs psycopg on `database_url_sync` (the comma-host libpq form), and its search_path handling needs verifying against the current multi-host code, which this branch predates.
- pgbouncer caveat. per-connection `SET search_path` and asyncpg `server_settings` are unreliable under transaction pooling. needs a documented note at minimum.
- in alembic `env.py` the `CREATE SCHEMA` / `SET search_path` use f-string interpolation instead of `Identifier`. it's operator-supplied config not user input so it's not a live injection, but it's inconsistent with the safer pattern you used in `database.py`. worth aligning.

real deployment feature with a sound core. rebase, add the env + docs entries, and sort the two-URL / pgbouncer edges and it's in good shape.

PR #335, opened 2026-04-30
feat(observability): propagate user-supplied run metadata to OTEL trace context

MarioAlessandroNapoli:
We're running aegra as the runtime for an enterprise agent platform we're building, and we hit this while wiring up Langfuse for our staging traces. Aegra accepts `metadata` on `POST /threads/{id}/runs` — it's part of `RunCreate` and the LangGraph SDK ships it as a first-class parameter — but the value is silently dropped before any span is emitted. The OTEL trace context that ends up in Langfuse/Phoenix/Generic exporters carries only the four runtime keys aegra injects internally (`run_id`, `thread_id`, `graph_id`, `original_request_id`) and nothing the caller asked to attach. From the observability side this makes filtering and correlation impossible without rewriting metadata server-side, which defeats the point of the field.

The data is lost in five places along the way. `RunCreate.metadata` reaches the API, but `_prepare_run` in `services/run_preparation.py` never reads it, so it doesn't make it into the `RunJob`. `RunJob` itself has no field for it, so even if `_prepare_run` did read the value there'd be nowhere to put it. The serialized `execution_params` JSONB column doesn't store it. The worker's `_restore_trace_context` builds a hardcoded four-key dict for `set_trace_context`. The local executor's `make_run_trace_context` does the same. Each step is small on its own; together they make the field non-functional.

This PR threads user metadata through the whole pipeline. We add a `run_metadata: dict[str, Any]` field on `RunJob` with a default of `{}` (backward compatible — old serialized rows missing the key deserialize fine). `_prepare_run` now reads `request.metadata or {}` and passes it to `RunJob`. `to_execution_params` and `from_run_orm` round-trip the field through JSONB. `make_run_trace_context` accepts an optional `extra_metadata` argument and merges it into the trace metadata dict. `_restore_trace_context` does the equivalent merge on the worker path.

When a user-supplied key collides with one of the four reserved system keys, we let the system value win and log a warning. The reasoning is that `run_id` and friends are runtime invariants — they are the join keys between logs, spans, and the `runs` table, and a caller spoofing them would silently break correlation across the platform. The warning gives the user a clear debug signal without breaking the request. We considered three alternatives (user wins, namespace separation under `aegra.*`, hard reject with 422); system-wins is the smallest viable change and the most reversible if community feedback prefers something else.

We kept the implementation vendor-neutral. `set_trace_context` already emits `langfuse.trace.metadata.<key>` for each metadata entry, and we preserved that pattern untouched, so Langfuse traces continue to expose metadata as queryable trace-level fields. Phoenix and Generic OTLP exporters surface the same attributes through standard span attribute mechanics. No Langfuse-specific imports leak into core code.

One note on values: `RunCreate.metadata` is typed `dict[str, Any]` (matching the LangGraph SDK), but OTEL span attributes only accept primitives (`str`, `int`, `float`, `bool`) and sequences of primitives. Non-primitive values (nested dicts, custom objects) reach `span.set_attribute` and are dropped with an OTEL warning rather than serialized. We considered validating at the `RunCreate` boundary but kept the SDK-aligned permissive type to avoid divergence; users pre-flatten or stringify complex values if they want them in trace metadata.

Tags are out of scope for this PR. The LangGraph SDK doesn't ship a `tags` parameter on `runs.create`, so adding one here would diverge from upstream contract. Users who need tag-like filtering can put a list under `metadata={"tags": [...]}` and Langfuse will surface it as `langfuse.trace.metadata.tags`. We can follow up with first-class tags if there's demand.

### Type of change

- [x] `feat`: new feature
- [ ] `fix`: bug fix (could also be argued — the field is documented but non-functional)

### Related issues

None open at the time of writing.

### Changes made

- `models/run_job.py`: added `run_metadata: dict[str, Any] = Field(default_factory=dict)` to `RunJob`. Updated `to_execution_params()` and `from_run_orm()` to round-trip the field with safe defaults
- `services/run_preparation.py`: `_prepare_run` now reads `request.metadata or {}` and passes it to `RunJob`
- `services/worker_executor.py`: `_restore_trace_context` merges `job.run_metadata` with the four system keys; system keys take precedence, collisions emit a `WARN` log
- `services/local_executor.py`: `submit` passes `job.run_metadata` to `make_run_trace_context`
- `observability/span_enrichment.py`: `make_run_trace_context` gains an optional `extra_metadata` parameter; system metadata still overrides on collision

### Testing

- [ ] E2E tests added/updated
- [ ] Manual testing performed — TBD before submission

Unit coverage:

- `tests/unit/test_models/test_run_job.py`: round-trip `run_metadata` through `to_execution_params` / `from_run_orm`, plus default-empty behavior for legacy rows
- `tests/unit/test_services/test_worker_executor.py`: `_restore_trace_context` merges user metadata, system keys override on collision, warning is logged
- `tests/unit/test_observability/test_span_enrichment.py`: `make_run_trace_context` includes `extra_metadata` and respects collision precedence
- `tests/unit/test_services/test_run_preparation.py` (extending existing if present, otherwise new): `RunCreate.metadata` flows into `RunJob.run_metadata`

### Validation in our staging environment

We're already running this fork against our staging Cloud Run deployment of aegra `0.9.6` (built from the wheel of this branch) and the change works end-to-end. We validated two paths.

**Happy path — propagation of legitimate user keys.** A real user query that dispatches a sub-agent emits two Langfuse traces, one for the main run and one for the worker run. The worker trace's `metadata` now contains the user-supplied keys we send from our orchestrator (`turn_id`, `subagent_type`, `workspace_id`, `job_id`, `parent_thread_id`) alongside the four reserved system keys (`run_id`, `thread_id`, `graph_id`, `original_request_id`). Phoenix aliases (`session.id` alongside `langfuse.session.id`) are emitted as expected, so the change is consumed correctly by both Langfuse and Phoenix-shaped exporters in our setup. In practice this lets us filter Langfuse traces by `metadata.turn_id` and reconstruct the full chain of a multi-agent turn (one main run + N worker dispatches) without joining against an external database — before this PR we maintained a SQL bridge for that.

**Anti-spoofing path — collision policy under attack.** Since our regular traffic doesn't reuse reserved key names, we ran a dedicated probe to exercise the merge policy directly. We POSTed a run with `metadata` containing the four reserved keys all set to obvious sentinel values (`SPOOFED-RUN-ID-AAAA`, `SPOOFED-THREAD-ID-BBBB`, `SPOOFED-GRAPH-CCCC`, `SPOOFED-CORR-DDDD`) plus two legitimate keys (`tenant`, `test_id`). On the resulting Langfuse trace, all four reserved fields hold the actual aegra runtime values (worker run UUID, real thread UUID, real graph name, real correlation_id) — none of the spoofed sentinels surface. The two legitimate keys pass through to `metadata.tenant` and `metadata.test_id` as expected. The worker logs emit the four corresponding `User metadata key 'X' overridden by system value` warnings from `merge_run_metadata`, one per reserved-key collision, confirming the policy fires audibly without breaking the request.

### Backward compatibility

The new `run_metadata` field defaults to `{}`. Rows persisted before this PR don't have the key in `execution_params`; `from_run_orm` falls back to `{}` so they deserialize without error. No migration is required because `execution_params` is JSONB.

### Vendor neutrality

We emit user metadata as `langfuse.trace.metadata.<key>` (existing convention). Phoenix and Generic OTLP exporters surface the same span attributes through standard OTEL mechanics — no exporter-specific code paths added.

### Additional notes

The four reserved keys (`run_id`, `thread_id`, `graph_id`, `original_request_id`) are runtime invariants and the join keys between logs, spans, and the `runs` table. Letting a caller override them would silently break correlation in any nontrivial setup, which is why the merge gives system precedence. The warning log makes the override visible to anyone debugging.

## Summary by CodeRabbit

* **New Features**
  * Runs accept optional custom run metadata preserved through serialization and reconstruction.
  * Run metadata is merged into trace contexts so user metadata appears alongside system runtime identifiers.
  * System-reserved keys win on collisions (user entries dropped with a warning); non-primitive user values are dropped with a warning.
  * original_request_id is included in metadata only when a correlation id is present.

* **Tests**
  * Added unit tests for metadata defaults, serialization, merging, collision handling, non-primitive dropping, and trace propagation.

<h3>Greptile Summary</h3>

This PR threads user-supplied `RunCreate.metadata` through the full execution pipeline so it appears in OTEL trace spans alongside the four runtime system keys. The change touches five layers: API-boundary validation, `RunJob` serialization, JSONB persistence, and both the local-executor and worker-executor trace-context restore paths.

- **`RunCreate` validator** (`runs.py`): a new `validate_metadata_shape` field validator enforces key regex, a 32-key cap, primitive-only values, and 512-char string cap — returning a clean 422 for violating payloads rather than silently dropping values downstream.
- **`merge_run_metadata`** (`span_enrichment.py`): a new helper merges user metadata with `system_metadata`; system keys always win and emit a warning on collision. By using the live `system_metadata` dict as the collision authority rather than a separate reserved-key constant, the implementation avoids the drift issue noted in prior review threads — including the `original_request_id` local-executor behaviour, which is now correctly gated on whether the runtime actually supplies that key.
- **`RunJob`** (`run_job.py`) + **JSONB round-trip**: `run_metadata` defaults to `{}` so pre-PR rows deserialize without a migration.

Safe to merge; all three concerns from previous review rounds are addressed and the implementation is internally consistent across both executor paths.

The propagation chain is complete and correctly wired end-to-end. The dynamic `system_metadata` approach eliminates the reserved-key drift concern from earlier threads. Validation at the `RunCreate` boundary ensures non-primitives and over-large payloads are rejected before they reach the OTEL layer. The only gap is a missing direct unit test for the `_prepare_run` → `RunJob.run_metadata` assignment, which the PR description committed to but omitted.

libs/aegra-api/src/aegra_api/services/run_preparation.py — the one-liner that starts the propagation chain has no dedicated test.

| Filename | Overview |
|----------|----------|
| libs/aegra-api/src/aegra_api/models/runs.py | Adds `validate_metadata_shape` field_validator to `RunCreate.metadata`, enforcing key regex, key count limit, primitive value types, and string length — all consistently matching the documented constraints in observability.mdx. |
| libs/aegra-api/src/aegra_api/observability/span_enrichment.py | Introduces `merge_run_metadata` helper and updates `make_run_trace_context` to accept `extra_metadata`; uses a dynamic `system_metadata` dict as the collision authority, resolving the previous `_RESERVED_METADATA_KEYS` drift concern. `original_request_id` is correctly absent from the local-executor `system_metadata`. |
| libs/aegra-api/src/aegra_api/models/run_job.py | Adds `run_metadata: dict[str, Any] = Field(default_factory=dict)` with round-trip support in `to_execution_params` and `from_run_orm`; the `or {}` fallback in `from_run_orm` correctly handles both missing keys (legacy rows) and JSONB null values. |
| libs/aegra-api/src/aegra_api/services/worker_executor.py | Conditionally gates `original_request_id` in `system_metadata` on the worker path, so collision detection and warning fire correctly when a user supplies that key with a real correlation ID present. |
| libs/aegra-api/src/aegra_api/services/run_preparation.py | One-line change passes `request.metadata or {}` to `RunJob.run_metadata`; straightforward and correct, though no dedicated `test_run_preparation.py` test was added (mentioned in the PR description but absent from the changeset). |
| libs/aegra-api/tests/unit/test_observability/test_span_enrichment.py | Adds unit tests for `merge_run_metadata` and two new `make_run_trace_context` tests; also includes a real-SDK end-to-end class (`TestSpanEnrichmentEndToEnd`) wiring up `TracerProvider`+`InMemorySpanExporter` to verify the full propagation chain without mocks. |

</details>

<h3>Sequence Diagram</h3>

```mermaid
sequenceDiagram
    participant Client
    participant API as POST /threads/{id}/runs
    participant RC as RunCreate (Pydantic)
    participant PR as _prepare_run
    participant RJ as RunJob
    participant JSONB as execution_params (JSONB)
    participant LE as LocalExecutor.submit
    participant WE as _restore_trace_context
    participant MRM as merge_run_metadata
    participant SC as set_trace_context

    Client->>API: metadata: {tenant, flag, ...}
    API->>RC: validate_metadata_shape()
    Note over RC: key regex, count ≤32, primitives only, str ≤512
    RC-->>API: 422 if invalid
    API->>PR: RunCreate (validated)
    PR->>RJ: run_metadata = request.metadata or {}
    RJ->>JSONB: to_execution_params() stores run_metadata

    alt Local executor path
        RJ->>LE: job.run_metadata
        LE->>MRM: extra_metadata=job.run_metadata, system={run_id,thread_id,graph_id}
        MRM-->>LE: merged (system wins on collision)
        LE->>SC: metadata=merged → contextvars.Context
    else Worker executor path
        JSONB->>RJ: from_run_orm() → run_metadata
        RJ->>WE: job.run_metadata
        WE->>MRM: extra_metadata=job.run_metadata, system={run_id,thread_id,graph_id[,original_request_id]}
        MRM-->>WE: merged (system wins on collision)
        WE->>SC: metadata=merged
    end
```

<a href="https://app.greptile.com/ide/claude-code?prompt=Fix%20the%20following%201%20code%20review%20issue.%20Work%20through%20them%20one%20at%20a%20time%2C%20proposing%20concise%20fixes.%0A%0A---%0A%0A%23%23%23%20Issue%201%20of%201%0Alibs%2Faegra-api%2Fsrc%2Faegra_api%2Fservices%2Frun_preparation.py%3A240-241%0A**Missing%20test%20for%20%60_prepare_run%60%20metadata%20propagation**%0A%0AThe%20PR%20description%20explicitly%20lists%20%60tests%2Funit%2Ftest_services%2Ftest_run_preparation.py%60%20as%20a%20testing%20target%20%E2%80%94%20%22extending%20existing%20if%20present%2C%20otherwise%20new%22%20%E2%80%94%20to%20verify%20%60RunCreate.metadata%60%20flows%20into%20%60RunJob.run_metadata%60.%20That%20file%20is%20absent%20from%20the%20changeset.%20The%20%60test_runs_wait.py%60%20update%20only%20adds%20%60request.metadata%20%3D%20None%60%20to%20fix%20the%20existing%20mock%3B%20it%20doesn't%20assert%20that%20a%20non-%60None%60%20metadata%20dict%20reaches%20the%20%60RunJob%60.%20Because%20this%20one-liner%20sits%20at%20the%20head%20of%20the%20propagation%20chain%2C%20a%20future%20accidental%20regression%20here%20%28e.g.%20forgetting%20%60or%20%7B%7D%60%2C%20renaming%20the%20field%29%20would%20produce%20no%20span%20attributes%20in%20Langfuse%2FPhoenix%20with%20no%20failing%20tests.%0A%0A&repo=aegra%2Faegra&pr=335&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=2">Fix All in Claude Code</picture></a> <a href="https://app.greptile.com/api/ide/cursor?prompt=Fix%20the%20following%201%20code%20review%20issue.%20Work%20through%20them%20one%20at%20a%20time%2C%20proposing%20concise%20fixes.%0A%0A---%0A%0A%23%23%23%20Issue%201%20of%201%0Alibs%2Faegra-api%2Fsrc%2Faegra_api%2Fservices%2Frun_preparation.py%3A240-241%0A**Missing%20test%20for%20%60_prepare_run%60%20metadata%20propagation**%0A%0AThe%20PR%20description%20explicitly%20lists%20%60tests%2Funit%2Ftest_services%2Ftest_run_preparation.py%60%20as%20a%20testing%20target%20%E2%80%94%20%22extending%20existing%20if%20present%2C%20otherwise%20new%22%20%E2%80%94%20to%20verify%20%60RunCreate.metadata%60%20flows%20into%20%60RunJob.run_metadata%60.%20That%20file%20is%20absent%20from%20the%20changeset.%20The%20%60test_runs_wait.py%60%20update%20only%20adds%20%60request.metadata%20%3D%20None%60%20to%20fix%20the%20existing%20mock%3B%20it%20doesn't%20assert%20that%20a%20non-%60None%60%20metadata%20dict%20reaches%20the%20%60RunJob%60.%20Because%20this%20one-liner%20sits%20at%20the%20head%20of%20the%20propagation%20chain%2C%20a%20future%20accidental%20regression%20here%20%28e.g.%20forgetting%20%60or%20%7B%7D%60%2C%20renaming%20the%20field%29%20would%20produce%20no%20span%20attributes%20in%20Langfuse%2FPhoenix%20with%20no%20failing%20tests.%0A%0A&pr=335&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCursorDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCursor.svg?v=2">Fix All in Cursor</picture></a> <a href="https://chatgpt.com/codex/deeplink?prompt=IMPORTANT%3A%20Work%20in%20the%20repository%20%22aegra%2Faegra%22%20on%20the%20existing%20branch%20%22feat%2Frun-metadata-propagation%22.%20Checkout%20that%20branch%20%E2%80%94%20do%20NOT%20create%20a%20new%20branch%20or%20open%20a%20new%20PR.%20Push%20your%20changes%20to%20%22feat%2Frun-metadata-propagation%22.%0A%0AFix%20the%20following%201%20code%20review%20issue.%20Work%20through%20them%20one%20at%20a%20time%2C%20proposing%20concise%20fixes.%0A%0A---%0A%0A%23%23%23%20Issue%201%20of%201%0Alibs%2Faegra-api%2Fsrc%2Faegra_api%2Fservices%2Frun_preparation.py%3A240-241%0A**Missing%20test%20for%20%60_prepare_run%60%20metadata%20propagation**%0A%0AThe%20PR%20description%20explicitly%20lists%20%60tests%2Funit%2Ftest_services%2Ftest_run_preparation.py%60%20as%20a%20testing%20target%20%E2%80%94%20%22extending%20existing%20if%20present%2C%20otherwise%20new%22%20%E2%80%94%20to%20verify%20%60RunCreate.metadata%60%20flows%20into%20%60RunJob.run_metadata%60.%20That%20file%20is%20absent%20from%20the%20changeset.%20The%20%60test_runs_wait.py%60%20update%20only%20adds%20%60request.metadata%20%3D%20None%60%20to%20fix%20the%20existing%20mock%3B%20it%20doesn't%20assert%20that%20a%20non-%60None%60%20metadata%20dict%20reaches%20the%20%60RunJob%60.%20Because%20this%20one-liner%20sits%20at%20the%20head%20of%20the%20propagation%20chain%2C%20a%20future%20accidental%20regression%20here%20%28e.g.%20forgetting%20%60or%20%7B%7D%60%2C%20renaming%20the%20field%29%20would%20produce%20no%20span%20attributes%20in%20Langfuse%2FPhoenix%20with%20no%20failing%20tests.%0A%0A"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCodexDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInCodex.svg?v=2">Fix All in Codex</picture></a>

<sub>Reviews (6): Last reviewed commit: ["fix(models): shorten run\_metadata commen..."](https://github.com/aegra/aegra/commit/9cac3b6c04be46e231e49b1c07075835f9124935) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30811794)</sub>

### DIFF
--- a/libs/aegra-api/src/aegra_api/models/run_job.py
+++ b/libs/aegra-api/src/aegra_api/models/run_job.py
@@ -67,6 +67,17 @@ class RunJob(BaseModel):
     user: User
     execution: RunExecution = RunExecution()
     behavior: RunBehavior = RunBehavior()
+    # User-supplied request metadata, persisted to ``execution_params``
+    # JSONB.  Validation is deliberately asymmetric: ``RunCreate.metadata``
+    # is strict at ingress (primitive-only union, key regex, length caps),
+    # while this field is permissive ``dict[str, Any]`` so legacy rows and
+    # any future schema drift remain readable on ``from_run_orm``.  Values
+    # may collide with reserved system keys (e.g. ``run_id``) — the row
+    # stores user input verbatim for audit fidelity, and the OTEL boundary
+    # in ``observability.span_enrichment.merge_run_metadata`` filters and
+    # warns at emit time.  Consume only via ``merge_run_metadata``; do not
+    # tighten this annotation without restoring an equivalent filter on
+    # read in ``from_run_orm``.
     run_metadata: dict[str, Any] = Field(default_factory=dict)
 
     def to_execution_params(self) -> dict[str, Any]:
--- a/libs/aegra-api/src/aegra_api/models/run_job.py
+++ b/libs/aegra-api/src/aegra_api/models/run_job.py
@@ -67,6 +67,7 @@ class RunJob(BaseModel):
     user: User
     execution: RunExecution = RunExecution()
     behavior: RunBehavior = RunBehavior()
+    run_metadata: dict[str, Any] = Field(default_factory=dict)
 
     def to_execution_params(self) -> dict[str, Any]:
         """Serialize for the ``execution_params`` JSONB column.
@@ -80,6 +81,7 @@ class RunJob(BaseModel):
             "user": self.user.model_dump(),
             "execution": self.execution.model_dump(),
             "behavior": self.behavior.model_dump(),
+            "run_metadata": self.run_metadata,
         }
 
     @classmethod
@@ -97,4 +99,5 @@ class RunJob(BaseModel):
             user=User.model_validate(params["user"]),
             execution=RunExecution.model_validate(params["execution"]),
             behavior=RunBehavior.model_validate(params["behavior"]),
+            run_metadata=params.get("run_metadata") or {},
         )
--- a/libs/aegra-api/src/aegra_api/models/run_job.py
+++ b/libs/aegra-api/src/aegra_api/models/run_job.py
@@ -67,17 +67,8 @@ class RunJob(BaseModel):
     user: User
     execution: RunExecution = RunExecution()
     behavior: RunBehavior = RunBehavior()
-    # User-supplied request metadata, persisted to ``execution_params``
-    # JSONB.  Validation is deliberately asymmetric: ``RunCreate.metadata``
-    # is strict at ingress (primitive-only union, key regex, length caps),
-    # while this field is permissive ``dict[str, Any]`` so legacy rows and
-    # any future schema drift remain readable on ``from_run_orm``.  Values
-    # may collide with reserved system keys (e.g. ``run_id``) — the row
-    # stores user input verbatim for audit fidelity, and the OTEL boundary
-    # in ``observability.span_enrichment.merge_run_metadata`` filters and
-    # warns at emit time.  Consume only via ``merge_run_metadata``; do not
-    # tighten this annotation without restoring an equivalent filter on
-    # read in ``from_run_orm``.
+    # JSONB-persisted; deliberately ``dict[str, Any]`` so legacy/drifted rows stay readable via
+    # ``from_run_orm``. OTEL boundary in ``merge_run_metadata`` filters at emit. Do not narrow.
     run_metadata: dict[str, Any] = Field(default_factory=dict)
 
     def to_execution_params(self) -> dict[str, Any]:
--- a/libs/aegra-api/src/aegra_api/models/runs.py
+++ b/libs/aegra-api/src/aegra_api/models/runs.py
@@ -79,11 +79,12 @@ class RunCreate(BaseModel):
 
     # Request metadata (top-level in payload).  Reaches OTEL trace
     # attributes as ``langfuse.trace.metadata.<key>`` (and the
-    # OpenInference ``metadata.<key>`` alias on Phoenix targets).  Values
-    # are restricted to OTEL-attribute primitives because the SDK silently
-    # drops nested values; surfacing the rejection as a 422 makes the
-    # contract honest in the OpenAPI schema.
-    metadata: dict[str, str | int | float | bool] | None = Field(
+    # OpenInference ``metadata.<key>`` alias on Phoenix targets).  The
+    # field is annotated ``dict[str, Any]`` rather than a primitive
+    # union so a malformed payload produces one actionable 422 message
+    # from ``validate_metadata_shape`` instead of N parallel union-arm
+    # errors (one per primitive type Pydantic tries) per offending key.
+    metadata: dict[str, Any] | None = Field(
         None,
         description=(
             "Request metadata propagated to OTEL trace attributes "
@@ -100,23 +101,29 @@ class RunCreate(BaseModel):
     @classmethod
     def validate_metadata_shape(
         cls,
-        v: dict[str, str | int | float | bool] | None,
-    ) -> dict[str, str | int | float | bool] | None:
-        """Enforce key shape, key count, and string-value length.
+        metadata: dict[str, Any] | None,
+    ) -> dict[str, Any] | None:
+        """Enforce key shape, key count, value type, and string-value length.
 
-        Pydantic's type annotation already rejects non-primitive values;
-        this validator covers the bounds the type system can't express.
+        Validation runs entirely here (rather than relying on a primitive
+        union on the field type) so each violation produces one clear
+        error message instead of N parallel union-arm errors per offending
+        key — easier for clients to surface to humans.
         """
-        if v is None:
+        if metadata is None:
             return None
-        if len(v) > _METADATA_MAX_KEYS:
-            raise ValueError(f"metadata exceeds {_METADATA_MAX_KEYS} keys (got {len(v)})")
-        for key, value in v.items():
+        if len(metadata) > _METADATA_MAX_KEYS:
+            raise ValueError(f"metadata exceeds {_METADATA_MAX_KEYS} keys (got {len(metadata)})")
+        for key, value in metadata.items():
             if not _METADATA_KEY_RE.match(key):
                 raise ValueError(f"metadata key {key!r} must match {_METADATA_KEY_RE.pattern}")
+            if not isinstance(value, (str, int, float, bool)):
+                raise ValueError(
+                    f"metadata value for key {key!r} must be str/int/float/bool, got {type(value).__name__}"
+                )
             if isinstance(value, str) and len(value) > _METADATA_MAX_VALUE_LEN:
                 raise ValueError(f"metadata value for key {key!r} exceeds {_METADATA_MAX_VALUE_LEN} characters")
-        return v
+        return metadata
 
     @model_validator(mode="after")
     def validate_input_command_exclusivity(self) -> Self:
--- a/libs/aegra-api/src/aegra_api/models/runs.py
+++ b/libs/aegra-api/src/aegra_api/models/runs.py
@@ -1,5 +1,6 @@
 """Run-related Pydantic models for Agent Protocol"""
 
+import re
 from datetime import datetime
 from typing import Any, Literal, Self
 
@@ -12,6 +13,18 @@ from pydantic import (
 )
 
 from aegra_api.utils.status_compat import validate_run_status
+
+# Constraints for ``RunCreate.metadata`` keys/values, enforced at request
+# time so the OpenAPI schema is honest about what reaches OTEL.  Without
+# these limits a tenant could submit thousands of keys, megabyte-scale
+# values, or nested structures — all of which would either be silently
+# dropped by ``merge_run_metadata`` or balloon span size past the OTEL
+# collector limits.  Bounds chosen to be generous for legitimate use
+# (tenant id, feature flag, environment, sub-agent type, ...) while
+# closing the DoS surface.
+_METADATA_KEY_RE = re.compile(r"^[A-Za-z0-9_-]{1,64}$")
+_METADATA_MAX_KEYS = 32
+_METADATA_MAX_VALUE_LEN = 512
 
 
 class RunCreate(BaseModel):
@@ -64,11 +77,46 @@ class RunCreate(BaseModel):
         description="Whether to include subgraph events in streaming. When True, includes events from all subgraphs. When False (default when None), excludes subgraph events. Defaults to False for backwards compatibility.",
     )
 
-    # Request metadata (top-level in payload)
-    metadata: dict[str, Any] | None = Field(
+    # Request metadata (top-level in payload).  Reaches OTEL trace
+    # attributes as ``langfuse.trace.metadata.<key>`` (and the
+    # OpenInference ``metadata.<key>`` alias on Phoenix targets).  Values
+    # are restricted to OTEL-attribute primitives because the SDK silently
+    # drops nested values; surfacing the rejection as a 422 makes the
+    # contract honest in the OpenAPI schema.
+    metadata: dict[str, str | int | float | bool] | None = Field(
         None,
-        description="Request metadata (e.g., from_studio flag)",
+        description=(
+            "Request metadata propagated to OTEL trace attributes "
+            "(``langfuse.trace.metadata.<key>``).  Keys must match "
+            "``[A-Za-z0-9_-]{1,64}``.  Values must be primitive "
+            "(``str``, ``int``, ``float``, ``bool``); string values are "
+            "capped at 512 characters.  Maximum 32 keys.  Use this for "
+            "filterable attributes (tenant, feature flag, environment, "
+            "sub-agent type) rather than payload data."
+        ),
     )
+
+    @field_validator("metadata", mode="after")
+    @classmethod
+    def validate_metadata_shape(
+        cls,
+        v: dict[str, str | int | float | bool] | None,
+    ) -> dict[str, str | int | float | bool] | None:
+        """Enforce key shape, key count, and string-value length.
+
+        Pydantic's type annotation already rejects non-primitive values;
+        this validator covers the bounds the type system can't express.
+        """
+        if v is None:
+            return None
+        if len(v) > _METADATA_MAX_KEYS:
+            raise ValueError(f"metadata exceeds {_METADATA_MAX_KEYS} keys (got {len(v)})")
+        for key, value in v.items():
+            if not _METADATA_KEY_RE.match(key):
+                raise ValueError(f"metadata key {key!r} must match {_METADATA_KEY_RE.pattern}")
+            if isinstance(value, str) and len(value) > _METADATA_MAX_VALUE_LEN:
+                raise ValueError(f"metadata value for key {key!r} exceeds {_METADATA_MAX_VALUE_LEN} characters")
+        return v
 
     @model_validator(mode="after")
     def validate_input_command_exclusivity(self) -> Self:
--- a/libs/aegra-api/src/aegra_api/observability/span_enrichment.py
+++ b/libs/aegra-api/src/aegra_api/observability/span_enrichment.py
@@ -22,6 +22,7 @@ Usage::
 
 import contextvars
 import logging
+from typing import Any
 
 from opentelemetry.context import Context
 from opentelemetry.sdk.trace import ReadableSpan, Span, SpanProcessor
@@ -31,7 +32,20 @@ logger = logging.getLogger(__name__)
 # Metadata keys reserved for runtime state; user-supplied values for
 # these keys are dropped (system value wins) so they remain reliable
 # join keys between logs, spans, and the runs table.
-_RESERVED_METADATA_KEYS = frozenset({"run_id", "thread_id", "graph_id", "original_request_id"})
+#
+# ``original_request_id`` is intentionally not reserved: only the worker
+# path injects it (when an HTTP correlation-id is present), and it is
+# absent on the local-executor path. Reserving it would make
+# ``merge_run_metadata`` drop a user-supplied value with a misleading
+# "overridden by system value" warning whenever the system value is
+# missing — a silent data loss with no guarantee benefit.
+_RESERVED_METADATA_KEYS = frozenset({"run_id", "thread_id", "graph_id"})
+
+# OTEL span attributes only accept primitive scalar types. The
+# observability SDK silently drops any other value at attribute-set
+# time, so we filter at this layer and emit an aegra-level warning
+# instead of letting drops happen invisibly inside the SDK.
+_PRIMITIVE_ATTR_TYPES: tuple[type, ...] = (str, int, float, bool)
 
 # Per-request context variable holding span attributes to inject.
 # None means no trace context is set; on_start() is a no-op in that case.
@@ -127,16 +141,22 @@ def set_trace_context(
 
 
 def merge_run_metadata(
-    extra_metadata: dict[str, str | int | float | bool] | None,
+    extra_metadata: dict[str, Any] | None,
     system_metadata: dict[str, str | int | float | bool],
 ) -> dict[str, str | int | float | bool]:
     """Merge user-supplied metadata with system-injected runtime keys.
 
-    The reserved keys (``run_id``, ``thread_id``, ``graph_id``,
-    ``original_request_id``) are runtime invariants and join keys between
-    logs, spans, and the runs table.  When a user-supplied key collides
-    with a reserved key, the system value wins and a warning is logged so
-    the override is visible during debugging without breaking the request.
+    The reserved keys (``run_id``, ``thread_id``, ``graph_id``) are
+    runtime invariants and join keys between logs, spans, and the runs
+    table. When a user-supplied key collides with a reserved key, the
+    system value wins and a warning is logged so the override is visible
+    during debugging without breaking the request.
+
+    Non-primitive values (anything other than ``str``, ``int``, ``float``,
+    ``bool``) are dropped with a warning. OTEL span attributes accept
+    only primitives; passing a nested dict or list to ``span.set_attribute``
+    is a silent no-op at the SDK level. Filtering here surfaces the drop
+    with the offending key so callers can fix the payload upstream.
     """
     if not extra_metadata:
         return dict(system_metadata)
@@ -147,6 +167,14 @@ def merge_run_metadata(
                 "User metadata key '%s' overridden by system value; reserved keys: %s",
                 key,
                 sorted(_RESERVED_METADATA_KEYS),
+            )
+            continue
+        if not isinstance(value, _PRIMITIVE_ATTR_TYPES):
+            logger.warning(
+                "User metadata key '%s' has non-primitive type %s; dropping "
+                "(OTEL attributes accept str/int/float/bool only)",
+                key,
+                type(value).__name__,
             )
             continue
         merged[key] = value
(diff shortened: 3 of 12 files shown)


MarioAlessandroNapoli:
Rebased onto `main` @ `fc8fb61` (v0.9.11) to pick up the merged checkpoint-fork work (#321 / #346 / #347). No conflicts — the rebase covered the same files (`models/run_job.py`, `services/run_preparation.py`) but in non-overlapping regions, so the diff between this branch and `origin/main` is unchanged at 9 files / +243 / -9. CodeRabbit's earlier nitpick (module-scope `logging` import + stricter `caplog` assertion in `test_no_warning_when_keys_do_not_collide`) is included in the rebased history. Full unit suite is green at 956 passed.

libs/aegra-api/src/aegra_api/models/runs.py
```diff
@@ -64,12 +77,47 @@ class RunCreate(BaseModel):
...
+            "filterable attributes (tenant, feature flag, environment, "
+            "sub-agent type) rather than payload data."
+        ),
     )
 
+    @field_validator("metadata", mode="after")
+    @classmethod
+    def validate_metadata_shape(
+        cls,
+        v: dict[str, str | int | float | bool] | None,
```
Witaly3: nit: rename `v` → `metadata`. The Pydantic-v1 `v` convention is folklore at this point, and it reads especially awkwardly here because there's already a `value` variable in the loop below. Renaming makes the body self-documenting:

  ```python
  def validate_metadata_shape(cls, metadata: ...) -> ...:
      if metadata is None:
          return None
      if len(metadata) > _METADATA_MAX_KEYS:
          ...
      for key, value in metadata.items():
          ...
```

File: libs/aegra-api/src/aegra_api/models/runs.py [hunk 2]
```diff
@@ -64,12 +77,47 @@ class RunCreate(BaseModel):
...
 
-    # Request metadata (top-level in payload)
-    metadata: dict[str, Any] | None = Field(
+    # Request metadata (top-level in payload).  Reaches OTEL trace
+    # attributes as ``langfuse.trace.metadata.<key>`` (and the
+    # OpenInference ``metadata.<key>`` alias on Phoenix targets).  Values
+    # are restricted to OTEL-attribute primitives because the SDK silently
+    # drops nested values; surfacing the rejection as a 422 makes the
+    # contract honest in the OpenAPI schema.
+    metadata: dict[str, str | int | float | bool] | None = Field(
```
Witaly3: The strict union type produces a noisy 422 for nested values. Sending `{"metadata": {"nested": {"x": 1}}}` returns 4 parallel errors (`string_type`, `int_type`, `float_type`, `bool_type`) instead of one human-readable message — Pydantic tries every union arm and reports each failure.

  The PR description motivates the type narrowing as "surfacing the rejection as a 422 makes the contract honest in the OpenAPI schema", but the current error response is harder to act on than a `dict[str, Any]` would be.

  Suggestion: keep the field as `dict[str, Any] | None` and move the primitive-type check into the existing validator, so the rejection becomes a single clear message:

  ```python
  metadata: dict[str, Any] | None = Field(None, ...)

  @field_validator("metadata", mode="after")
  @classmethod
  def validate_metadata_shape(cls, metadata):
      if metadata is None:
          return None
      ...
      for key, value in metadata.items():
          if not isinstance(value, (str, int, float, bool)):
              raise ValueError(
                  f"metadata value for key {key!r} must be str/int/float/bool, "
                  f"got {type(value).__name__}"
              )
          ...
```

  OpenAPI loses the per-value type constraint, but the 422 message becomes actionable for the client.

File: libs/aegra-api/src/aegra_api/observability/span_enrichment.py [hunk 3]
```diff
@@ -118,24 +140,91 @@ def set_trace_context(
...
+    ``bool``) are dropped with a warning. OTEL span attributes accept
+    only primitives; passing a nested dict or list to ``span.set_attribute``
+    is a silent no-op at the SDK level. Filtering here surfaces the drop
+    with the offending key so callers can fix the payload upstream.
+    """
+    if not extra_metadata:
+        return dict(system_metadata)
+    merged: dict[str, str | int | float | bool] = {}
+    for key, value in extra_metadata.items():
+        if key in _RESERVED_METADATA_KEYS:
```
Witaly3: Both current callers (`make_run_trace_context`, `_restore_trace_context`) put `run_id`/`thread_id`/`graph_id` into `system_metadata` unconditionally, so this branch and the `key in system_metadata` branch below never disagree on outcome — only on warning text. The reserved set is effectively a slightly-more-informative warning, not a separate guarantee.

  That's not wrong, but it adds a second source of truth that can drift: if someone later adds a key to `_RESERVED_METADATA_KEYS` without also putting it in `system_metadata` for every caller, this branch silently drops user input with no system value to fall back on — exactly the failure mode the docstring at lines 36-41 calls out for `original_request_id`.

  Two options:
  1. Drop `_RESERVED_METADATA_KEYS` entirely; rely solely on `system_metadata` membership. One warning path, no drift surface.
  2. Keep the set, but treat it as a caller contract — assert at module-load or in a unit test that every reserved key is provided by every caller. That makes the redundancy a real safety net rather than dead code.

libs/aegra-api/src/aegra_api/services/run_preparation.py
```diff
@@ -237,6 +237,7 @@ async def _prepare_run(
             multitask_strategy=request.multitask_strategy,
             subgraphs=request.stream_subgraphs or False,
         ),
+        run_metadata=request.metadata or {},
```
Witaly3, line 240: A user value that collides with a reserved key (e.g. `metadata={"run_id": "spoofed"}`) is persisted verbatim in `execution_params.run_metadata`; the drop only happens at OTEL emit time inside `merge_run_metadata`. Verified live: the row contains `{"run_id": "spoofed", "tenant": "acme"}` even though the span carries the real `run_id`.

  That's defensible (audit fidelity, no data loss on persistence), but it sets a trap for any future reader of `run.execution_params.run_metadata` who reasonably assumes the field is post-filtered. Worth a one-line note on `RunJob.run_metadata` documenting the invariant: "may contain user-supplied collisions with reserved keys; consume only via `merge_run_metadata`."

libs/aegra-api/src/aegra_api/models/run_job.py
```diff
@@ -67,6 +67,7 @@ class RunJob(BaseModel):
     user: User
     execution: RunExecution = RunExecution()
     behavior: RunBehavior = RunBehavior()
+    run_metadata: dict[str, Any] = Field(default_factory=dict)
```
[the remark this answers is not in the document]
Witaly3: Re: narrowing `RunJob.run_metadata` to `dict[str, str | int | float | bool]` — I'd argue against this part of the suggestion. The primitive filter at the OTEL boundary is already in place (see `merge_run_metadata` lines 185-192, added in `73bdbbe`), so the runtime concern is covered.

  The wider `dict[str, Any]` on `RunJob` is load-bearing for `from_run_orm`: it deserializes from a JSONB column, where legacy rows (pre-PR) and any future schema drift can carry values that don't fit the narrower union. A narrowed annotation would either:
  - make `model_validate` raise on legacy/edited rows that were previously readable, or
  - require us to filter on read in `from_run_orm` to keep the type honest, duplicating logic that already exists at the OTEL emit boundary.

  The asymmetry between `RunCreate.metadata` (strict, validated at request time) and `RunJob.run_metadata` (defensive, validated at consumption time) is deliberate: ingress is strict, internal storage is permissive, the OTEL boundary filters again. Worth a one-line note on the field documenting this so it doesn't get "tightened" later, but the type itself is correct as-is.

Witaly3 (changes requested):
Functionally LGTM — verified locally:
  - `1251` unit/integration + `92` new tests pass
  - ruff + ty clean on changed files (the one ty diagnostic on `worker_executor.py:255` is pre-existing on `main`)
  - live e2e against `docker compose`: valid metadata persists in `execution_params.run_metadata`; nested/dotted/oversized payloads return 422; spoofed `run_id` triggers the expected warning from `merge_run_metadata` and the system value reaches the span attribute.

  Two non-inline points:

  **Docs gap.** The PR introduces a new public API contract (max 32 keys, key regex, 512-char string cap, primitive-only values, reserved-key anti-spoofing). Per the project guideline that user-facing changes ship with docs in the same PR, this needs a section in `docs/` covering the constraints and the propagation semantics (`langfuse.trace.metadata.<key>` on the root span). Worth doing before merge — it's the only thing I'd block on.

  **Phoenix metadata alias (out of scope, flagging as follow-up).** `set_trace_context` emits both `langfuse.user.id` and `user.id` for first-class fields, but metadata only goes through the `langfuse.trace.metadata.<key>` channel. On Phoenix targets that means user metadata reaches the span but isn't surfaced as Phoenix-native filterable metadata. Pre-existing limitation, not introduced here, and the PR's "vendor neutral, preserve existing convention" stance is defensible. Worth a follow-up issue, not a change request.

  The inline comments cover smaller polish items (validator parameter naming, error-message UX, redundant reserved-key branch, `RunJob.run_metadata` typing note). Nothing blocking other than the docs.

MarioAlessandroNapoli:
Thanks for the thorough review — all five points addressed in d718566.

**Inline polish**
- `runs.py:103` — validator parameter renamed `v → metadata`. Inner loop now uses the outer name without shadowing.
- `runs.py:86` — relaxed the field annotation back to `dict[str, Any] | None` and moved the primitive-type check into `validate_metadata_shape`. A nested-value payload now returns one actionable `422` (`metadata value for key 'k' must be str/int/float/bool, got dict`) instead of four parallel union-arm errors. We agree the per-value-type narrowing in OpenAPI was buying less than the noisy 422 was costing.
- `span_enrichment.py:172` — dropped `_RESERVED_METADATA_KEYS` entirely. `system_metadata` membership is now the single source of truth for "runtime-owned keys", which closes the drift surface you flagged. Existing tests cover the unified path with no behaviour change.
- `run_job.py:70` + `run_preparation.py:240` — added a docstring on `RunJob.run_metadata` documenting both the asymmetry (strict at `RunCreate.metadata` ingress, permissive `dict[str, Any]` for the JSONB column so legacy and forward-drifted rows stay readable on `from_run_orm`) and the collision invariant (rows may persist user-supplied keys that shadow reserved names; consumers must go through `merge_run_metadata`). One canonical note on the field rather than two scattered comments.

**Docs (the blocker)**
New \`Run metadata\` section in \`docs/guides/observability.mdx\` covering: the request shape, the constraints table (32 keys / regex / 512-char / primitive-only), the propagation channel (\`langfuse.trace.metadata.<key>\` on the root span), and the system-key collision rule.

**Phoenix metadata alias follow-up**
Opening a separate issue for the Phoenix-native metadata alias rather than expanding the scope of #335. Will link it here once filed.

Locally green: 978 unit tests pass, ruff format/lint clean, \`ty\` clean on changed files.

MarioAlessandroNapoli:
Phoenix metadata alias follow-up filed: #350. Captures the dual-attribute approach (the same pattern `set_trace_context` already uses for `user.id` and `session.id`) so the OpenInference `metadata.<key>` alias lands on the root span alongside `langfuse.trace.metadata.<key>`. Open for discussion there before any PR.

Witaly3 (changes requested):
@MarioAlessandroNapoli Agreed with CodeRabbit's withdrawal — the layering is the right call.

  One last polish on the same site, `libs/aegra-api/src/aegra_api/models run_job.py:70-79` (the comment block above `run_metadata`): the current note runs to 10 lines, which is over the project's 2-3 line cap on inline comments (per CLAUDE.md, "Code Comments (STRICT)"). CodeRabbit's suggested 3-liner above captures the same invariants — deliberate `dict[str, Any]`, JSONB legacy reason, OTEL boundary filters, "do not narrow" directive. Could you swap to that or an equivalent short form? Once it lands I'll approve.

MarioAlessandroNapoli:
Thanks for the patience on the polish rounds.

`9cac3b6` collapses the comment block to two lines with the four invariants kept: deliberate `dict[str, Any]`, JSONB legacy/forward-drift compat, `merge_run_metadata` filter at the OTEL boundary, and "do not narrow". The 2–3 line cap is on our pre-flight checklist now.

Witaly3 (approved):

ibbybuilds (approved):
nice work! LGTM

Issue #10, opened 2025-08-31
[FEATURE] Assistant versioning


arthiondaena:
**Is your feature request related to a problem? Please describe.**
The feature is already available in langgraph platform, but not present in aegra.

**Describe the solution you'd like**
Assistants to support versioning to track changes over time. Once you’ve created an assistant, subsequent edits to that assistant will create new versions.

**Use Case**
- Any application that lets user access all the assistant versions.
- Allows the user to change the version of the assistant that is to be used.

**Changes**
- API changes needed
New API routes include PATCH `/assistants/{assistant_id}` (for modifying the assistant), POST `/assistants/{assistant_id}/versions` (to fetch all the assistant versions) and POST `/assistants/{assistant_id}/latest` (to change the default version for the assistant)
- Database schema changes
New table named `versions` to store the version history of assistants and also new column in assistant column to store the assistant version that is currently being used.

**Priority**

- [ ] Low - Nice to have
- [x] Medium - Would significantly improve workflow
- [ ] High - Blocking current work
- [ ] Critical - Required for production use

PR #349, opened 2026-05-05
feat(api): add POST /threads/{id}/copy endpoint

MarioAlessandroNapoli:
We're running aegra as the runtime for an enterprise agent platform we're building, and we hit this while migrating from LangSmith Deployments. The official LangGraph SDK exposes `client.threads.copy(thread_id)` which calls `POST /threads/{id}/copy`, and our codebase already uses it to fork conversation state for branching/replay flows. Aegra `0.9.11` doesn't implement that endpoint, so the SDK call returns a 404 even though every other thread method works.

Before writing anything we ran the call against a live LangSmith Deployments instance to capture the real contract — neither the SDK nor the public docs spell it out fully. The request is `POST /threads/{id}/copy` with body `null`, and the response is HTTP 200 carrying the **full new Thread JSON** (the SDK type hint says `-> None`, but the server actually returns a populated object — we believe the SDK just discards it). The new thread gets a server-generated UUID, deep-copies `metadata`, `config.configurable`, `status`, `error`, and `values` from the source, regenerates `state_updated_at` to the time of the copy, and — crucially — its checkpoint history shares the **same `checkpoint_id` and `parent_checkpoint_id` graph** as the source. We verified this by listing `/threads/{src}/history` and `/threads/{tgt}/history` for the same source-target pair and observing identical `(checkpoint_id, parent_checkpoint_id)` tuples. That last point told us LangSmith implements this as a SQL-level copy of the checkpoint tables rather than as an `aupdate_state` replay (which would regenerate IDs).

This PR ports the same semantics to aegra. The endpoint lives in `api/threads.py:copy_thread` and the heavy lifting is in a new `services/thread_copy.py:copy_thread_atomically` helper. The helper opens a single connection on the existing `langgraph-checkpoint-postgres` pool, starts one Postgres transaction, INSERTs the new thread row, and then runs `INSERT INTO ... SELECT ... WHERE thread_id = %s` against `checkpoints`, `checkpoint_writes`, and `checkpoint_blobs` — substituting only the `thread_id` and preserving every other column verbatim. Going through one pool/one transaction makes the copy atomic: either the new thread row and its full checkpoint history land together, or nothing does. Doing the thread-row INSERT through the same pool (rather than through the SQLAlchemy session) was a deliberate choice — we tried the cross-pool variant first and ended up with a window where the checkpointer pool had committed checkpoint rows while the SQLAlchemy session had not yet committed the thread row, which would leave orphaned checkpoint rows on a partial failure.

Column lists for the three checkpoint tables are introspected at runtime via `information_schema.columns` rather than hard-coded. We did this specifically because `langgraph-checkpoint-postgres` has been adding columns over time (`task_path` was added in a recent migration), and pinning to a static column list would silently drop any future additions on copy. The introspection uses dict-style row access (`r["column_name"]`) because the aegra pool is configured with `row_factory=dict_row` in `core/database.py:50` — getting that right matters because `r[0]` raises `KeyError(0)` against a `dict_row` cursor, and we only caught it through end-to-end testing on a real Postgres rather than through the unit tests, which had been mocking the cursor.

Authorization reuses the `threads:create` event because copy is a creation operation from the caller's point of view. Source-thread ownership is verified before the copy — the helper only fires after `session.scalar(select(ThreadORM).where(... user_id == user.identity))` confirms the source belongs to the caller, so a caller cannot copy someone else's thread.

We're aware of the recently-merged checkpoint-fork work in #321 / #346 / #347. That work is complementary, not overlapping: those PRs let you fork-resume **the same thread** at a chosen checkpoint via `POST /runs` with a `checkpoint` payload, while this endpoint creates a **new thread** with the full source history attached. Both code paths can coexist; nothing here changes the input/state-update flow they fixed.

### Type of change

- [x] `feat`: new feature

### Related issues / PRs

None open at the time of writing. Complementary to merged #321 / #346 / #347 (checkpoint-based fork-resume).

### Changes made

- `api/threads.py`: new `POST /threads/{thread_id}/copy` endpoint. 404 when the source thread is missing or owned by another user; 200 + full Thread JSON otherwise.
- `services/thread_copy.py` (new): `copy_thread_atomically` helper plus internal `_copy_checkpoint_table` and `_table_columns` used by it. Single transaction on the langgraph-checkpoint-postgres pool covers the thread row INSERT and the three checkpoint table copies.
- `tests/unit/test_services/test_thread_copy.py` (new): 7 unit tests covering column introspection (dict-row access matching the pool factory), INSERT...SELECT shape, defensive skip when `thread_id` column is missing from the schema, atomic helper sequencing (thread row before checkpoint copy, all on the same connection), JSON serialization of metadata, and error handling when the pool is uninitialized.

### Testing

- [x] Unit tests added/updated — 7 new tests, full unit suite green at 956 passed locally
- [ ] E2E tests added/updated
- [x] Manual testing performed — see "Validation in our staging environment" below

### Validation in our staging environment

We deployed this branch to our staging Cloud Run instance (built from a wheel of this branch) and verified the endpoint end-to-end against a real Postgres backend.

We created a source thread via `POST /threads`, ran a small graph against it via `POST /runs/wait` to populate two checkpoints (`step=-1 source=input`, `step=0 source=loop`), then issued `POST /threads/{src}/copy`. The response came back HTTP 200 with the full new Thread JSON; the new thread inherited `metadata`, `status`, and the populated `values` from the source as expected. We then listed `/threads/{tgt}/history` and compared `(checkpoint_id, parent_checkpoint_id)` tuples against the source — both threads had identical pairs, confirming the SQL-level copy preserves the checkpoint graph (whereas an `aupdate_state` replay would have regenerated the IDs). We also ran the same probe with a deliberately failing graph to confirm the source's `error`-status is copied through, mirroring the LangSmith reference behaviour.

### Backward compatibility

Additive endpoint — no existing routes touched. No schema changes; the helper uses the existing `thread` table and the langgraph-checkpoint-postgres tables that aegra already provisions. Threads created before this PR can be passed as the source; rows missing optional columns deserialize fine because the column list is introspected per-call.

### Additional notes

The behaviour around `status=src.status` is intentional: if the source has `status='running'` or `status='error'`, the new thread inherits that without an actual run attached. This matches what we observed on LangSmith Deployments when the source had a failed run, and we kept it this way for drop-in compatibility — flagging it explicitly because it is mildly surprising at first glance.

## Summary by CodeRabbit

* **New Features**
  * Added a "duplicate thread" endpoint to atomically copy a user-owned thread into a new thread_id, preserving conversation history, checkpoints, status, and metadata (owner updated), and returning the new thread.

* **Bug Fixes / Reliability**
  * Ensures atomic, repeatable-copy behavior and emits structured audit logging; returns 404 if source missing and 500 on unexpected failures.

* **Tests**
  * Added unit and integration tests for copy semantics, SQL safety, metadata handling, and error cases.

<h3>Greptile Summary</h3>

This PR implements `POST /threads/{thread_id}/copy`, a new endpoint that atomically deep-copies a thread and its full checkpoint history to a new `thread_id`, matching `client.threads.copy()` semantics from LangSmith Deployments. The heavy lifting is in `services/thread_copy.py`, which opens a single REPEATABLE READ transaction on the `lg_pool` and uses `INSERT … SELECT` with `psycopg.sql.Identifier`-composed identifiers and `pg_catalog.pg_attribute`+`to_regclass` for live column introspection.

- `api/threads.py`: adds `copy_thread` endpoint; auth-handler metadata mutations are captured and forwarded, and `threads:create` fires with both the new and source thread IDs in the payload.
- `services/thread_copy.py`: new `copy_thread_atomically` helper with `_copy_thread_row` (overrides `thread_id`, `user_id`, timestamps, and active-status reset) and `_copy_checkpoint_table` (verbatim copy of all three checkpoint tables, preserving the `checkpoint_id` chain).
- Tests cover unit, integration, and E2E levels; docs are updated with a new \"Copying threads\" section.

Additive endpoint with no changes to existing routes; atomic cross-pool copy is well-tested at unit, integration, and E2E levels.

All previously flagged issues (silent skip, SQL identifier injection, schema-path mismatch, discarded handler return, wrong auth-event payload) are correctly addressed in this revision. The remaining notes are style and hygiene items that do not affect correctness.

The `session.commit()` + reload block in `api/threads.py` sits outside the try/except after a cross-pool commit — worth a second look to confirm the duplicate-copy scenario is acceptable for the deployment's retry posture.

| Filename | Overview |
|----------|----------|
| libs/aegra-api/src/aegra_api/api/threads.py | Adds POST /threads/{thread_id}/copy endpoint. Auth-handler metadata capture and new-thread-id payload are correct. The post-commit reload sits outside the try/except, creating a narrow window where a durable copy can return HTTP 500. |
| libs/aegra-api/src/aegra_api/services/thread_copy.py | New service. Single REPEATABLE READ transaction on lg_pool, column introspection via pg_attribute+to_regclass, psycopg.sql.Identifier throughout — all previously flagged issues resolved. |
| libs/aegra-api/tests/unit/test_services/test_thread_copy.py | 7 unit tests covering dict-row access, INSERT…SELECT shape, missing-column raise, REPEATABLE READ sequencing, metadata merge order, and pool-not-initialized guard. |
| libs/aegra-api/tests/integration/test_api/test_threads_crud.py | Adds two integration tests for handler metadata propagation and consistent thread_id across the auth event and persistence call. |
| libs/aegra-api/tests/e2e/test_threads/test_thread_copy.py | Three E2E tests: empty-thread copy, checkpoint-chain preservation, and 404 on missing source. |
| docs/guides/threads-and-state.mdx | Documentation accurately describes copy semantics, atomicity guarantee, and authorization model. |

<h3>Sequence Diagram</h3>

<a href="#gh-light-mode-only">

```mermaid
sequenceDiagram
    participant C as Client
    participant E as copy_thread endpoint
    participant A as handle_event (auth)
    participant S as SQLAlchemy session
    participant P as lg_pool (psycopg)
    participant DB as Postgres

    C->>E: "POST /threads/{id}/copy"
    E->>A: "threads:create event {thread_id: new_id, src_thread_id}"
    A-->>E: filters (metadata overrides)
    E->>S: "SELECT thread WHERE id=src AND user_id=caller"
    S->>DB: query
    DB-->>S: src row (or 404)
    S-->>E: src ORM object

    Note over E,P: copy_thread_atomically — single REPEATABLE READ tx
    E->>P: connection()
    P->>DB: BEGIN / SET TRANSACTION ISOLATION LEVEL REPEATABLE READ
    P->>DB: pg_attribute introspection (thread + 3 checkpoint tables)
    P->>DB: "INSERT INTO thread ... SELECT ... WHERE id=src"
    P->>DB: "INSERT INTO checkpoints ... SELECT ... WHERE thread_id=src"
    P->>DB: "INSERT INTO checkpoint_writes ... SELECT ... WHERE thread_id=src"
    P->>DB: "INSERT INTO checkpoint_blobs ... SELECT ... WHERE thread_id=src"
    P->>DB: COMMIT
    DB-->>P: committed

    E->>S: session.commit() (refresh snapshot)
    E->>S: "SELECT thread WHERE id=new_id"
    S->>DB: query
    DB-->>S: new thread row
    S-->>E: new ORM object
    E-->>C: 200 Thread JSON
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
%%{init: {'theme': 'base', 'themeVariables': {"darkMode": true, "background": "#0d1117", "primaryColor": "#21262d", "primaryTextColor": "#e6edf3", "primaryBorderColor": "#8b949e", "lineColor": "#8b949e", "textColor": "#e6edf3", "edgeLabelBackground": "#161b22", "actorBkg": "#21262d", "actorBorder": "#8b949e", "actorTextColor": "#e6edf3", "actorLineColor": "#8b949e", "signalColor": "#8b949e", "signalTextColor": "#e6edf3", "noteBkgColor": "#373320", "noteBorderColor": "#d4a72c", "noteTextColor": "#f0e6c0", "labelBoxBkgColor": "#21262d", "labelBoxBorderColor": "#8b949e", "labelTextColor": "#e6edf3", "loopTextColor": "#e6edf3", "activationBkgColor": "#30363d", "activationBorderColor": "#8b949e"}}}%%
sequenceDiagram
    participant C as Client
    participant E as copy_thread endpoint
    participant A as handle_event (auth)
    participant S as SQLAlchemy session
    participant P as lg_pool (psycopg)
    participant DB as Postgres

    C->>E: "POST /threads/{id}/copy"
    E->>A: "threads:create event {thread_id: new_id, src_thread_id}"
    A-->>E: filters (metadata overrides)
    E->>S: "SELECT thread WHERE id=src AND user_id=caller"
    S->>DB: query
    DB-->>S: src row (or 404)
    S-->>E: src ORM object

    Note over E,P: copy_thread_atomically — single REPEATABLE READ tx
    E->>P: connection()
    P->>DB: BEGIN / SET TRANSACTION ISOLATION LEVEL REPEATABLE READ
    P->>DB: pg_attribute introspection (thread + 3 checkpoint tables)
    P->>DB: "INSERT INTO thread ... SELECT ... WHERE id=src"
    P->>DB: "INSERT INTO checkpoints ... SELECT ... WHERE thread_id=src"
    P->>DB: "INSERT INTO checkpoint_writes ... SELECT ... WHERE thread_id=src"
    P->>DB: "INSERT INTO checkpoint_blobs ... SELECT ... WHERE thread_id=src"
    P->>DB: COMMIT
    DB-->>P: committed

    E->>S: session.commit() (refresh snapshot)
    E->>S: "SELECT thread WHERE id=new_id"
    S->>DB: query
    DB-->>S: new thread row
    S-->>E: new ORM object
    E-->>C: 200 Thread JSON
```

</a>

<details open><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `libs/aegra-api/src/aegra_api/api/threads.py`, line 43-56 ([link](https://github.com/aegra/aegra/blob/9a7ca7bce7af529041a54f061140abb65eddcfb0/libs/aegra-api/src/aegra_api/api/threads.py#L43-L56)) 

   <a href="#">P1</a> **`handle_event` fires before the source ownership check**

   `handle_event` is called at line 44–47 before the `select(ThreadORM).where(... user_id == user.identity)` guard at line 49–54. If the source thread doesn't exist or belongs to another user the endpoint raises 404 — but the handler has already run. Any handler that increments a quota counter, creates a billing record, or provisions a tenant slot on the new `thread_id` will leave those side-effects in place even though no thread ever lands in the database.

   The fix is straightforward: move the ownership check to the top (before `new_id = str(uuid4())`), or at minimum before the `handle_event` call. `new_id` can be generated after the guard passes.

   <a href="https://app.greptile.com/ide/claude-code?prompt=This%20is%20a%20comment%20left%20during%20a%20code%20review.%0APath%3A%20libs%2Faegra-api%2Fsrc%2Faegra_api%2Fapi%2Fthreads.py%0ALine%3A%2043-56%0A%0AComment%3A%0A**%60handle_event%60%20fires%20before%20the%20source%20ownership%20check**%0A%0A%60handle_event%60%20is%20called%20at%20line%2044%E2%80%9347%20before%20the%20%60select%28ThreadORM%29.where%28...%20user_id%20%3D%3D%20user.identity%29%60%20guard%20at%20line%2049%E2%80%9354.%20If%20the%20source%20thread%20doesn't%20exist%20or%20belongs%20to%20another%20user%20the%20endpoint%20raises%20404%20%E2%80%94%20but%20the%20handler%20has%20already%20run.%20Any%20handler%20that%20increments%20a%20quota%20counter%2C%20creates%20a%20billing%20record%2C%20or%20provisions%20a%20tenant%20slot%20on%20the%20new%20%60thread_id%60%20will%20leave%20those%20side-effects%20in%20place%20even%20though%20no%20thread%20ever%20lands%20in%20the%20database.%0A%0AThe%20fix%20is%20straightforward%3A%20move%20the%20ownership%20check%20to%20the%20top%20%28before%20%60new_id%20%3D%20str%28uuid4%28%29%29%60%29%2C%20or%20at%20minimum%20before%20the%20%60handle_event%60%20call.%20%60new_id%60%20can%20be%20generated%20after%20the%20guard%20passes.%0A%0AHow%20can%20I%20resolve%20this%3F%20If%20you%20propose%20a%20fix%2C%20please%20make%20it%20concise.&repo=aegra%2Faegra&pr=349&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixInClaudeDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixInClaude.svg?v=2">Fix in Claude Code</picture></a> <a href="https://app.greptile.com/api/ide/cursor?prompt=This%20is%20a%20comment%20left%20during%20a%20code%20review.%0APath%3A%20libs%2Faegra-api%2Fsrc%2Faegra_api%2Fapi%2Fthreads.py%0ALine%3A%2043-56%0A%0AComment%3A%0A**%60handle_event%60%20fires%20before%20the%20source%20ownership%20check**%0A%0A%60handle_event%60%20is%20called%20at%20line%2044%E2%80%9347%20before%20the%20%60select%28ThreadORM%29.where%28...%20user_id%20%3D%3D%20user.identity%29%60%20guard%20at%20line%2049%E2%80%9354.%20If%20the%20source%20thread%20doesn't%20exist%20or%20belongs%20to%20another%20user%20the%20endpoint%20raises%20404%20%E2%80%94%20but%20the%20handler%20has%20already%20run.%20Any%20handler%20that%20increments%20a%20quota%20counter%2C%20creates%20a%20billing%20record%2C%20or%20provisions%20a%20tenant%20slot%20on%20the%20new%20%60thread_id%60%20will%20leave%20those%20side-effects%20in%20place%20even%20though%20no%20thread%20ever%20lands%20in%20the%20database.%0A%0AThe%20fix%20is%20straightforward%3A%20move%20the%20ownership%20check%20to%20the%20top%20%28before%20%60new_id%20%3D%20str%28uuid4%28%29%29%60%29%2C%20or%20at%20minimum%20before%20the%20%60handle_event%60%20call.%20%60new_id%60%20can%20be%20generated%20after%20the%20guard%20passes.%0A%0AHow%20can%20I%20resolve%20this%3F%20If%20you%20propose%20a%20fix%2C%20please%20make%20it%20concise.&pr=349&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixInCursorDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixInCursor.svg?v=2">Fix in Cursor</picture></a> <a href="https://chatgpt.com/codex/deeplink?prompt=IMPORTANT%3A%20Work%20in%20the%20repository%20%22aegra%2Faegra%22%20on%20the%20existing%20branch%20%22feat%2Fthread-copy%22.%20Checkout%20that%20branch%20%E2%80%94%20do%20NOT%20create%20a%20new%20branch%20or%20open%20a%20new%20PR.%20Push%20your%20changes%20to%20%22feat%2Fthread-copy%22.%0A%0AThis%20is%20a%20comment%20left%20during%20a%20code%20review.%0APath%3A%20libs%2Faegra-api%2Fsrc%2Faegra_api%2Fapi%2Fthreads.py%0ALine%3A%2043-56%0A%0AComment%3A%0A**%60handle_event%60%20fires%20before%20the%20source%20ownership%20check**%0A%0A%60handle_event%60%20is%20called%20at%20line%2044%E2%80%9347%20before%20the%20%60select%28ThreadORM%29.where%28...%20user_id%20%3D%3D%20user.identity%29%60%20guard%20at%20line%2049%E2%80%9354.%20If%20the%20source%20thread%20doesn't%20exist%20or%20belongs%20to%20another%20user%20the%20endpoint%20raises%20404%20%E2%80%94%20but%20the%20handler%20has%20already%20run.%20Any%20handler%20that%20increments%20a%20quota%20counter%2C%20creates%20a%20billing%20record%2C%20or%20provisions%20a%20tenant%20slot%20on%20the%20new%20%60thread_id%60%20will%20leave%20those%20side-effects%20in%20place%20even%20though%20no%20thread%20ever%20lands%20in%20the%20database.%0A%0AThe%20fix%20is%20straightforward%3A%20move%20the%20ownership%20check%20to%20the%20top%20%28before%20%60new_id%20%3D%20str%28uuid4%28%29%29%60%29%2C%20or%20at%20minimum%20before%20the%20%60handle_event%60%20call.%20%60new_id%60%20can%20be%20generated%20after%20the%20guard%20passes.%0A%0AHow%20can%20I%20resolve%20this%3F%20If%20you%20propose%20a%20fix%2C%20please%20make%20it%20concise."><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixInCodexDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixInCodex.svg?v=2">Fix in Codex</picture></a>
</details>

<sub>Reviews (9): Last reviewed commit: ["fix(api): harden thread copy per review ..."](https://github.com/aegra/aegra/commit/8cad03857241919b91725850fffff9dc23f7c83b) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=30812020)</sub>

**Context used:**

- Context used - CLAUDE.md ([source](https://app.greptile.com/aegra/github/aegra/aegra/-/custom-context?memory=1640ab19-5310-4fc5-aaf6-3760087c66e8))

### DIFF
--- a/docs/guides/threads-and-state.mdx
+++ b/docs/guides/threads-and-state.mdx
@@ -152,6 +152,35 @@ Deleting a thread cancels any active runs and removes all state:
 await client.threads.delete(thread_id)
 ```
 
+## Copying threads
+
+`POST /threads/{thread_id}/copy` produces a new thread whose checkpoint history is identical to the source, matching the `client.threads.copy()` semantics on LangSmith Deployments. The `checkpoint_id` and `parent_checkpoint_id` chain are preserved verbatim, so the new thread can be queried by the same checkpoint identifiers as the source and `get_state_at_checkpoint(...)` returns the same values on both sides.
+
+```python
+new_thread = await client.threads.copy(thread_id)
+print(new_thread["thread_id"])  # freshly generated UUID
+```
+
+### What is preserved
+
+- The full checkpoint history (`checkpoints`, `checkpoint_writes`, `checkpoint_blobs` rows) under the new `thread_id`.
+- Thread `status` from the source — including non-idle states like `running` or `error` if present at the time of the copy.
+- Thread metadata, with two adjustments described below.
+
+### What is regenerated
+
+- `thread_id`: a fresh UUID is always generated server-side.
+- `created_at` and `updated_at`: set to the time of the copy, not inherited from the source.
+- `metadata.owner`: rewritten to the caller's identity, matching `POST /threads` (creation) behaviour. The source thread's owner attribution is intentionally not inherited so any code reading `metadata.owner` instead of the canonical `user_id` column stays consistent.
+
+### Atomicity
+
+Both the new thread row and the three checkpoint-table `INSERT...SELECT` statements run inside a single Postgres transaction at `REPEATABLE READ` isolation on the checkpointer pool. A concurrent writer on the source thread (for example a run mid-execution) cannot leave the new thread with a partially-copied checkpoint history: the snapshot is pinned at transaction start, and on any failure the whole copy is rolled back. There is no intermediate state where a new thread row exists without its checkpoint chain or vice versa.
+
+### Authorization
+
+The endpoint requires the caller to own the source thread (`user_id == user.identity`); a missing or non-owned source thread returns `404`. Custom auth handlers registered for `threads:create` see the new `thread_id` in the event payload (alongside `src_thread_id`) so they can provision per-thread resources keyed to the new row.
+
 ## Thread metadata
 
 Threads automatically track metadata about their usage:
--- a/docs/guides/threads-and-state.mdx
+++ b/docs/guides/threads-and-state.mdx
@@ -164,22 +164,23 @@ print(new_thread["thread_id"])  # freshly generated UUID
 ### What is preserved
 
 - The full checkpoint history (`checkpoints`, `checkpoint_writes`, `checkpoint_blobs` rows) under the new `thread_id`.
-- Thread `status` from the source — including non-idle states like `running` or `error` if present at the time of the copy.
-- Thread metadata, with two adjustments described below.
+- Thread metadata, inherited as-is except for the two adjustments below. This includes keys written by previous runs, auth handlers, or admin tooling — if a key must not propagate to copies, strip it in a `threads:create` handler.
+- `assistant_id` / `graph_id` in metadata are kept intentionally, so a copy can continue the same conversation (fork use case).
 
 ### What is regenerated
 
 - `thread_id`: a fresh UUID is always generated server-side.
 - `created_at` and `updated_at`: set to the time of the copy, not inherited from the source.
+- `status`: active states (`busy`/`running`) reset to `idle`; other states (`idle`, `error`, `interrupted`) copy verbatim. The copy has no `runs` row, so an inherited active status would never transition and the thread would appear permanently stuck.
 - `metadata.owner`: rewritten to the caller's identity, matching `POST /threads` (creation) behaviour. The source thread's owner attribution is intentionally not inherited so any code reading `metadata.owner` instead of the canonical `user_id` column stays consistent.
 
 ### Atomicity
 
-Both the new thread row and the three checkpoint-table `INSERT...SELECT` statements run inside a single Postgres transaction at `REPEATABLE READ` isolation on the checkpointer pool. A concurrent writer on the source thread (for example a run mid-execution) cannot leave the new thread with a partially-copied checkpoint history: the snapshot is pinned at transaction start, and on any failure the whole copy is rolled back. There is no intermediate state where a new thread row exists without its checkpoint chain or vice versa.
+The new thread row and the three checkpoint-table `INSERT...SELECT` statements all run inside a single Postgres transaction at `REPEATABLE READ` isolation on the checkpointer pool, reading the source thread row and its checkpoint history from one pinned snapshot. A concurrent writer on the source (for example a run mid-execution) cannot leave the copy with a partially-copied history or a thread row inconsistent with it: the snapshot is fixed at transaction start, and any failure rolls the whole copy back. There is no intermediate state where a new thread row exists without its checkpoint chain or vice versa.
 
 ### Authorization
 
-The endpoint requires the caller to own the source thread (`user_id == user.identity`); a missing or non-owned source thread returns `404`. Custom auth handlers registered for `threads:create` see the new `thread_id` in the event payload (alongside `src_thread_id`) so they can provision per-thread resources keyed to the new row.
+The endpoint requires the caller to own the source thread (`user_id == user.identity`); a missing or non-owned source thread returns `404`. Copies reuse the `threads:create` auth event rather than a dedicated `threads:copy` hook: handlers see the new `thread_id` alongside `src_thread_id` in the payload, so a handler that needs copy-specific policy (blocking cross-tenant copies, a separate quota counter) can branch on the presence of `src_thread_id`.
 
 ## Thread metadata
 
--- a/libs/aegra-api/src/aegra_api/api/threads.py
+++ b/libs/aegra-api/src/aegra_api/api/threads.py
@@ -890,13 +890,21 @@ async def copy_thread(
     # Atomic INSERT thread row + checkpoint INSERT...SELECT inside a single
     # Postgres transaction on the checkpointer pool. Avoids the cross-pool
     # split that would leave orphaned checkpoint rows on partial failure.
-    await copy_thread_atomically(
-        src_thread_id=thread_id,
-        new_thread_id=new_id,
-        src_status=src.status,
-        src_metadata=src.metadata_json or {},
-        user_identity=user.identity,
-    )
+    try:
+        await copy_thread_atomically(
+            src_thread_id=thread_id,
+            new_thread_id=new_id,
+            src_status=src.status,
+            src_metadata=src.metadata_json or {},
+            user_identity=user.identity,
+        )
+    except Exception:
+        logger.exception(
+            "Failed to copy thread",
+            src_thread_id=thread_id,
+            new_thread_id=new_id,
+        )
+        raise HTTPException(500, "Failed to copy thread") from None
 
     # Reload the freshly-inserted row through the ORM so the response is
     # built from the same shape ``_serialize_thread`` expects elsewhere.
--- a/libs/aegra-api/src/aegra_api/api/threads.py
+++ b/libs/aegra-api/src/aegra_api/api/threads.py
@@ -906,6 +906,14 @@ async def copy_thread(
         )
         raise HTTPException(500, "Failed to copy thread") from None
 
+    # End the implicit session transaction before reloading. Postgres' default
+    # READ COMMITTED snapshot would let the lg_pool INSERT show through anyway,
+    # but stricter isolation levels (REPEATABLE READ / SERIALIZABLE) freeze the
+    # session snapshot at transaction start and the new row would be invisible.
+    # Committing here makes the reload robust regardless of the engine's
+    # isolation_level setting.
+    await session.commit()
+
     # Reload the freshly-inserted row through the ORM so the response is
     # built from the same shape ``_serialize_thread`` expects elsewhere.
     new_thread = await session.scalar(select(ThreadORM).where(ThreadORM.thread_id == new_id))
--- a/libs/aegra-api/src/aegra_api/api/threads.py
+++ b/libs/aegra-api/src/aegra_api/api/threads.py
@@ -873,8 +873,19 @@ async def copy_thread(
     are regenerated to the time of the copy.
     """
     # Authorization check — modelled on POST /threads (creation event).
+    # We generate ``new_id`` up-front and pass it inside the event payload
+    # so handlers that provision per-thread resources (tenant_id, quota
+    # markers, billing tags) key those resources to the new row rather
+    # than the source.  We also capture the handler return value the same
+    # way ``create_thread`` does, so handler-enforced metadata mutations
+    # propagate to the copy — without that capture, ``/copy`` would be a
+    # bypass path for invariants every directly-created thread carries.
+    new_id = str(uuid4())
     ctx = build_auth_context(user, "threads", "create")
-    await handle_event(ctx, {"thread_id": thread_id})
+    filters = await handle_event(
+        ctx,
+        {"thread_id": new_id, "src_thread_id": thread_id},
+    )
 
     src = await session.scalar(
         select(ThreadORM).where(
@@ -885,7 +896,16 @@ async def copy_thread(
     if not src:
         raise HTTPException(404, f"Thread '{thread_id}' not found")
 
-    new_id = str(uuid4())
+    # Apply handler-returned metadata mutations on top of the source's
+    # metadata before the copy lands.  Mirrors ``create_thread`` lines
+    # 172–180.  ``copy_thread_atomically`` then rewrites
+    # ``metadata["owner"] = user.identity`` last, so the security
+    # invariant wins over any owner field a handler might inject.
+    src_metadata = dict(src.metadata_json or {})
+    if filters and "metadata" in filters:
+        handler_meta = filters["metadata"]
+        if isinstance(handler_meta, dict):
+            src_metadata = {**src_metadata, **handler_meta}
 
     # Atomic INSERT thread row + checkpoint INSERT...SELECT inside a single
     # Postgres transaction on the checkpointer pool. Avoids the cross-pool
@@ -895,7 +915,7 @@ async def copy_thread(
             src_thread_id=thread_id,
             new_thread_id=new_id,
             src_status=src.status,
-            src_metadata=src.metadata_json or {},
+            src_metadata=src_metadata,
             user_identity=user.identity,
         )
     except Exception:
--- a/libs/aegra-api/src/aegra_api/api/threads.py
+++ b/libs/aegra-api/src/aegra_api/api/threads.py
@@ -861,25 +861,13 @@ async def copy_thread(
     user: User = Depends(get_current_user),
     session: AsyncSession = Depends(get_session),
 ) -> Thread:
-    """Deep-copy a thread, preserving its checkpoint history.
+    """Deep-copy a thread, preserving its ``checkpoint_id`` chain.
 
-    Creates a new thread with a freshly generated ``thread_id`` whose
-    checkpoint chain is identical to the source: ``checkpoint_id`` and
-    ``parent_checkpoint_id`` are kept as-is so the new thread can be
-    queried by the same checkpoint identifiers as the source. Thread-level
-    metadata and status are deep-copied as-is — including statuses such
-    as ``running`` or ``error`` if present on the source — matching
-    LangSmith Deployments behaviour. ``created_at`` and ``updated_at``
-    are regenerated to the time of the copy.
+    Status and metadata are inherited from the source (including ``running``
+    or ``error``); ``created_at``/``updated_at`` are regenerated.
     """
-    # Authorization check — modelled on POST /threads (creation event).
-    # We generate ``new_id`` up-front and pass it inside the event payload
-    # so handlers that provision per-thread resources (tenant_id, quota
-    # markers, billing tags) key those resources to the new row rather
-    # than the source.  We also capture the handler return value the same
-    # way ``create_thread`` does, so handler-enforced metadata mutations
-    # propagate to the copy — without that capture, ``/copy`` would be a
-    # bypass path for invariants every directly-created thread carries.
+    # ``new_id`` is generated up-front and exposed to the auth event so handlers
+    # can key per-thread resources (tenant, quota, billing) to the new row.
     new_id = str(uuid4())
     ctx = build_auth_context(user, "threads", "create")
     filters = await handle_event(
@@ -896,20 +884,14 @@ async def copy_thread(
     if not src:
         raise HTTPException(404, f"Thread '{thread_id}' not found")
 
-    # Apply handler-returned metadata mutations on top of the source's
-    # metadata before the copy lands.  Mirrors ``create_thread`` lines
-    # 172–180.  ``copy_thread_atomically`` then rewrites
-    # ``metadata["owner"] = user.identity`` last, so the security
-    # invariant wins over any owner field a handler might inject.
+    # Mirror ``create_thread``: handler metadata wins over source, then
+    # ``copy_thread_atomically`` rewrites ``owner`` last as security invariant.
     src_metadata = dict(src.metadata_json or {})
     if filters and "metadata" in filters:
         handler_meta = filters["metadata"]
         if isinstance(handler_meta, dict):
             src_metadata = {**src_metadata, **handler_meta}
 
-    # Atomic INSERT thread row + checkpoint INSERT...SELECT inside a single
-    # Postgres transaction on the checkpointer pool. Avoids the cross-pool
-    # split that would leave orphaned checkpoint rows on partial failure.
     try:
         await copy_thread_atomically(
             src_thread_id=thread_id,
@@ -926,19 +908,13 @@ async def copy_thread(
         )
         raise HTTPException(500, "Failed to copy thread") from None
 
-    # End the implicit session transaction before reloading. Postgres' default
-    # READ COMMITTED snapshot would let the lg_pool INSERT show through anyway,
-    # but stricter isolation levels (REPEATABLE READ / SERIALIZABLE) freeze the
-    # session snapshot at transaction start and the new row would be invisible.
-    # Committing here makes the reload robust regardless of the engine's
-    # isolation_level setting.
+    # Commit before reload so REPEATABLE READ/SERIALIZABLE engines see the row
+    # (the lg_pool INSERT is invisible to a frozen session snapshot otherwise).
     await session.commit()
 
-    # Reload the freshly-inserted row through the ORM so the response is
-    # built from the same shape ``_serialize_thread`` expects elsewhere.
     new_thread = await session.scalar(select(ThreadORM).where(ThreadORM.thread_id == new_id))
     if new_thread is None:
-        # Should be unreachable: the atomic INSERT just committed.
+        # Unreachable: the atomic INSERT just committed.
         raise HTTPException(500, "Thread copy committed but the row could not be reloaded")
 
     return _serialize_thread(new_thread)
--- a/libs/aegra-api/src/aegra_api/api/threads.py
+++ b/libs/aegra-api/src/aegra_api/api/threads.py
@@ -35,6 +35,7 @@ from aegra_api.models import (
 )
 from aegra_api.models.errors import CONFLICT, NOT_FOUND
 from aegra_api.services.streaming_service import streaming_service
+from aegra_api.services.thread_copy import copy_thread_atomically
 from aegra_api.services.thread_state_service import ThreadStateService
 from aegra_api.utils.run_utils import strip_pinned_config_keys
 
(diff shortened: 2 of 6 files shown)


MarioAlessandroNapoli:
Self-compliance pass before maintainer review — pushed in 7022b54 against CLAUDE.md / CONTRIBUTING.md. No API behaviour change.

**Versioning** — bumped `aegra-api` and `aegra-cli` to `0.9.12` (non-breaking new feature → patch). `aegra-api~=0.9.0` constraint already permits it; `uv.lock` updated for the two version strings only.

**Comments** — collapsed all `# `-prefixed blocks and docstrings in `services/thread_copy.py` and `api/threads.py copy_thread` to the 2–3 line cap. Kept the non-obvious WHYs (REPEATABLE READ snapshot pin, `metadata.owner` rewrite as security invariant, `dict_row` factory contract, the `session.commit()` before reload that defends stricter isolation levels).

**Signature** — `copy_thread_atomically` now has the `*` kw-only separator (5 named params). Call site already uses kwargs.

**Docs** — new \`Copying threads\` section in \`docs/guides/threads-and-state.mdx\`: request shape, what is preserved/regenerated, atomicity guarantee, authorization model.

**E2E** — \`tests/e2e/test_threads/test_thread_copy.py\` against a real server:
- empty-thread copy → fresh \`thread_id\`, inherited \`status\`
- run-bearing copy → identical \`checkpoint_id\` chain (length + per-row identity via \`client.threads.get_history()\`)
- missing / non-owned source → \`404\`

Local: ruff clean, ty delta zero vs \`main\`, 63 existing unit + integration tests on \`thread_copy.py\` still pass.

Witaly3 (changes requested):
.

Witaly3 (changes requested):
I bench-tested the PR end-to-end against a freshly-built `aegra` container (`docker compose up` on the PR branch) with direct SQL access to the live schema. Unit + integration tests pass; two E2E pass; the third E2E (`test_thread_copy_preserves_checkpoint_chain`) requires a working LLM run and was not exercised locally — it should pass in CI where the key is set, see point 9 below for a complementary suggestion that does not rely on a real run.

### What the live tests confirm

- **Checkpoint chain copy works.** Seeded 3 checkpoints + matching writes + blobs on the source via raw SQL; after `POST /threads/{id}/copy`, all three tables had identical row counts under the new `thread_id`, and the `checkpoint_id` chain was preserved verbatim.
- **Atomicity inside `copy_thread_atomically` is real.** Installed a `BEFORE INSERT` trigger on `checkpoint_writes` that raises mid-loop, fired `/copy`, observed `HTTP 500`. Counts of `thread`, `checkpoints`, `checkpoint_writes`, `checkpoint_blobs` were identical before and after — zero ghost rows. The rationale for writing `thread` through `lg_pool` (cross-table atomicity) is load-bearing, not arbitrary.

### Findings

#### 1. `services/thread_copy.py:95-100` — `thread` schema is duplicated, not the pool choice

After the live atomicity test above, writing the `thread` row through `lg_pool` is justified — refactoring it out to SQLAlchemy would split the transaction across two pools and reintroduce a real ghost-row failure mode. **Do not** restructure that.

The narrower issue is that the column list `("thread_id", "status", "metadata_json", "user_id", "created_at", "updated_at")` is hardcoded, while checkpoint tables are introspected via
`pg_attribute`. Asymmetric and fragile: a NOT NULL column added to `ThreadORM` (e.g. `tenant_id`, `deleted_at`) breaks this INSERT silently at runtime.

Two clean fixes that keep the lg_pool tx intact:
1. Introspect `thread` columns through `pg_attribute` the same way checkpoint tables are.
2. Build the INSERT off `ThreadORM.__table__.columns` at runtime — one source of truth with the ORM.

#### 2. `services/thread_copy.py:1-9` + `docs/guides/threads-and-state.mdx:178` — atomicity wording is too strong

`src.metadata_json` and `src.status` are read through the SQLAlchemy session *before* the lg_pool transaction opens (`api/threads.py:878-902`). The REPEATABLE READ snapshot pins only the
checkpoint INSERT…SELECT phase. A concurrent run committing metadata or status changes between the SQLAlchemy SELECT and the lg_pool BEGIN will produce a copy whose thread row is stale relative to the (consistent) checkpoint history. Window is short but real.

Either tighten the implementation (read `src.status` / `src.metadata_json` inside the lg_pool tx) or soften the docs:

> "best-effort consistent snapshot of checkpoint history; source thread row fields are read on a separate connection and may lag by milliseconds under concurrent writes"

#### 3. `api/threads.py:895-913` — two-pool invariant should be pinned in a comment

`copy_thread_atomically` commits via `lg_pool`. The subsequent `session.commit()` runs on the SQLAlchemy pool — by then `INSERT INTO thread` is already durable. Today this is harmless because the session has only the ownership SELECT. A future change adding `session.add(audit_log)` between these lines would introduce a real ghost-row bug (lg_pool committed, ORM rolled back).

Please add a short comment to the endpoint stating the invariant:

> *"After `copy_thread_atomically` returns, the session must hold no pending writes — the lg_pool tx is the source of truth. Do not `session.add(...)` below this line."*

#### 4. `services/thread_copy.py:93-94` — use the idiomatic isolation-level setter

```python
from psycopg import IsolationLevel

async with db_manager.lg_pool.connection() as conn:
    async with conn.transaction(isolation_level=IsolationLevel.REPEATABLE_READ):
        ...
```

`SET TRANSACTION` is honored only when no DML has run yet in the current tx. The current code works because it's the first statement after `conn.transaction()` opens BEGIN, but any future refactor prepending a guard SELECT, a logging query, or even a `SELECT 1` will silently demote the isolation level back to READ COMMITTED without any error. The `isolation_level=` kwarg makes the property structural rather than positional.

#### 5. `tests/unit/test_services/test_thread_copy.py` — rollback path is asserted, not tested

`_make_atomic_pool_mocks` wires `transaction()` as a context manager that always returns cleanly. There is no test that raises mid-loop through a checkpoint INSERT and asserts no partial state is observable afterwards.

Live-verified externally with this minimal repro (sharing for completeness — worth porting into the suite):

```python
# 1. Seed src thread + 3 checkpoints via direct SQL
# 2. CREATE TRIGGER fault_t BEFORE INSERT ON checkpoint_writes
#    FOR EACH ROW EXECUTE FUNCTION raise_exception()
# 3. POST /threads/{src_id}/copy → expect 500
# 4. Assert row count delta is 0 across thread, checkpoints,
#    checkpoint_writes, checkpoint_blobs
```

For a service whose selling point is atomicity, this should live in `tests/integration/` (real DB) — either with a testcontainers fixture or using the existing E2E DB connection with a
fault-injection trigger.

#### 6. `services/thread_copy.py:89-90` — document metadata propagation

`metadata["owner"] = user_identity` is the only rewrite — every other key in source metadata is inherited as-is, including any value written by previous runs, by handlers, or by admin tooling. This is structurally the same as `create_thread`, except there the metadata was authored by the caller; here it was authored by whatever code touched the source.

Add to `docs/guides/threads-and-state.mdx` § "What is preserved":

> "All source metadata except `owner` is inherited as-is, including keys written by previous runs, handlers, or admin tooling. If a key should not propagate to copies, strip it explicitly via a `threads:create` auth handler."

Also confirm in the `copy_thread_atomically` docstring that preserving `assistant_id` / `graph_id` is intentional (it is, for fork-conversation use cases) — distinguishes intent from oversight.

#### 7. `api/threads.py:872` — `threads:create` vs `threads:copy`

`threads:create` fires for copy and the payload includes `src_thread_id` as a hint. A handler that needs different policy for copies (no cross-tenant copy, separate quota counter, etc.) has to inspect that key — there is no dedicated hook.

Two defensible choices, please pick one explicitly:
1. **Keep `threads:create`**, document in the auth-handler reference that `src_thread_id` in the payload signals a copy. Cheaper, less API surface.
2. **Add a dedicated `threads:copy` event** that falls through to `threads:create` if no handler is registered. Strictly more flexible.

Failure mode is leaving it implicit — the next contributor with a copy-specific policy will not know where to hook in.

#### 8. `api/threads.py:903-909` — `from None` is inconsistent with the rest of the file

Sibling endpoints use `from e`:
- `get_thread_state` → `api/threads.py:412`
- `get_thread_state_at_checkpoint` → `api/threads.py:637`
- `update_thread_state` → `api/threads.py:556`

`logger.exception` already captures the traceback to logs — but `from None` also drops `__cause__` for any error-handling middleware downstream (Sentry, OTLP error tagging, custom 500 middleware). 
Live-verified: the fault-injection 500 surfaced as `{'error': 'internal_error', 'message': 'Failed to copy thread', 'details': None}` with no cause chain available to downstream middleware.

If the intent was to prevent SQL state from leaking client-side, that is error-middleware's responsibility, not the raise site. Switch to:

```python
raise HTTPException(500, "Failed to copy thread") from e
```

#### 9. `tests/e2e/test_threads/test_thread_copy.py` — coverage gaps

Two complementary gaps:

1. **`test_thread_copy_preserves_checkpoint_chain` couples checkpoint copy semantics to a live LLM run.** If the upstream provider throttles, geo-blocks, or changes response shape, the test fails on grounds unrelated to the feature. Consider an alternate path that seeds checkpoints via `client.threads.update_state(...)` (no real LLM needed) to assert chain preservation. Keep the LLM-driven variant for end-to-end confidence, add a seed-based variant for stability.

2. **Concurrency E2E missing.** The "REPEATABLE READ pinned against concurrent writers" claim is verified only by code reading. Test pattern: start a streaming run on `src_id`, fire
`client.threads.copy(src_id)` mid-stream, await both, assert checkpoint count on the copy matches exactly the count at the moment the copy transaction started (not the later, growing count on the source).

#### 10. `services/thread_copy.py` + `api/threads.py:866-868` — `busy` / `running` status copied without a `runs` row [LIVE-CONFIRMED]

Reproduced on a fresh container:

```
forced src.status = 'busy' via direct UPDATE
POST /threads/{src_id}/copy → 200
copy.status = 'busy'
runs rows for copy.thread_id = 0
```

`runs` has FK on `thread.thread_id` and is intentionally not copied. The copy inherits `status='busy'` from the source — no `runs` row will ever transition it out of `busy` because none was
created. The thread is permanently stuck.

Concrete consequences:
- `runs.list(copy_id)` returns empty while `thread.get(copy_id).status` says `busy` — silent divergence between two endpoints.
- Anything in UI / SDK that gates user actions on `status != "busy"` (e.g. "wait for current run to finish before starting a new one") will block forever on the copy.
- If `checkpoint_writes` had pending writes from the in-flight run on the source, those copy too — the next run on the copy may try to apply them to whatever channel they reference.

Two ways out, in order of preference:
1. **Force `status="idle"` for the copy** when source status is `busy` / `running`. Document the override in § "What is regenerated". Strongly recommended — copying a busy thread to get a
permanently-stuck copy is never what the caller wants.
2. Document explicitly that copying a busy thread produces an inconsistent copy and make it the caller's responsibility. Less safe.

Add an integration / E2E test asserting `status == "idle"` for a copy taken while the source is `busy` once a fix is chosen.

---

### Summary

Architecture (single lg_pool tx) is sound and atomicity is real — verified under fault injection. Main blockers I would not merge without:
- (10) `busy`-status data-corruption bug
- (1) hardcoded `thread` column list
- (2) atomicity wording fix in docs/docstring

The rest is polish but worth landing in the same PR.

Witaly3:
@MarioAlessandroNapoli hi, see my suggestions above, thank you

ibbybuilds (changes requested):
this is close @MarioAlessandroNapoli, nice work. the copy is atomic, the checkpoint chain (including parent pointers) is preserved, and the ownership handling is correct: source lookup filters on `user_id == user.identity` and 404s on missing-or-not-owned, and the new row is stamped with the caller's id and owner metadata. no cross-tenant path, which is the thing i most wanted to check on a copy endpoint. tests are the best of the current open PRs, unit + integration + e2e all present.

blockers before merge:

1. two forbidden references need to come out. `thread_copy.py` docstring says "forward-compat with upstream column additions", and `test_thread_copy.py` points at `infra/aegra/README.md`. we don't reference that internal package or private infra paths anywhere in the tree. reword both.
2. rebase, it's conflicting against main.

small stuff:
- drop the version bump (0.9.12). versions get set at merge time, and it'll collide with current main anyway.
- the `except Exception` around `copy_thread_atomically` in `threads.py` is a bit broad. it's logged and re-raised so it's not swallowing, but a narrower catch would be cleaner.

strip the two references and rebase and this is basically an approve. it's a solid LangSmith-parity feature and i want it in.

MarioAlessandroNapoli:
Thanks both — really thorough reviews, and apologies for the long gap on our side. Everything below landed in `8cad038`, rebased clean onto current `main` (only the version files conflicted).

The headline change: the thread row is now copied via an introspected `INSERT ... SELECT` from the source, inside the same `REPEATABLE READ` transaction as the checkpoint tables. That resolves three findings at once — the hardcoded column list, the busy-status corruption, and the stale-read window — so we've grouped by finding rather than by reviewer.

**Blockers**

- **(Witaly3 #10) busy/running copies into a stuck thread** — `services/thread_copy.py`: `status` is now `CASE WHEN status IN ('busy','running') THEN 'idle' ELSE status END`. The copy has no `runs` row, so an active status could never transition; terminal states copy verbatim. Regression test asserts the CASE in the generated SQL.
- **(Witaly3 #1) hardcoded `thread` column list** — the column list is introspected via `pg_attribute` (same path as the checkpoint tables) and the INSERT is `INSERT ... SELECT`, so any column we don't explicitly override — a future `tenant_id`, `deleted_at` — is inherited from the source instead of dropped. `test_inherits_unknown_future_column_verbatim` locks this.
- **(Witaly3 #2) atomicity wording** — resolved by tightening, not softening: `src.status`/`src.metadata_json` are no longer read through the SQLAlchemy session before the tx; they're read inside the pinned snapshot by the `INSERT ... SELECT`. Docs updated accordingly.
- **(ibby) two internal references** — dropped from the module docstring and the unit-test docstring.
- **(ibby) rebase / version bump** — rebased onto current `main`; version stays at `main`'s.

**Polish**

- **(#3) two-pool invariant** — `api/threads.py`: comment at the copy call site pinning "no `session.add(...)` after `copy_thread_atomically` — the lg_pool tx is the source of truth."
- **(#6) metadata propagation** — documented in `threads-and-state.mdx`: all source metadata except `owner` is inherited; `assistant_id`/`graph_id` kept intentionally for the fork case.
- **(#7) `threads:create` vs `threads:copy`** — picked option 1: kept `threads:create`, and documented (endpoint comment + Authorization section) that `src_thread_id` in the payload signals a copy so a handler can branch on it.
- **(#8) `from None` -> `from e`** — done, matching the sibling handlers (`get_thread_state:412`, `update_thread_state:556`).
- **(#4) isolation level** — we kept `SET TRANSACTION ... REPEATABLE READ` rather than the kwarg. psycopg 3.2.5's `AsyncConnection.transaction()` signature is `(savepoint_name, force_rollback)` — no `isolation_level` kwarg — and setting `conn.isolation_level` on a pooled connection leaks the level to the next borrower. `SET TRANSACTION` as the first in-tx statement is scoped to this transaction only; we reinforced the "must be first" comment to guard the ordering.
- **(ibby) `except Exception` breadth** — kept but switched to `from e`. It mirrors every sibling handler in the file (`except HTTPException: raise` / `except Exception as e: logger.exception(...); raise HTTPException(500, ...) from e`); the local catch buys the structured `src`/`new` log line and a preserved cause chain. Happy to narrow if you'd rather it diverge from the file pattern.

**Follow-up (#460)**

- **(#5, #9) real-Postgres test hardening for `/copy`** — the fault-injection rollback test, the seed-based (LLM-free) chain variant, and the concurrency test all need a real-DB harness with fault-injection triggers that the current unit/integration layers don't have. We've captured them in #460 rather than bolt a half-harness onto this PR. At unit level we did add the zero-affected-rows rollback path (source deleted mid-copy -> raise -> tx rolls back).

Local validation: `1742 passed` (unit + integration), ruff format/lint clean, `ty` delta 0 on the touched files.

PR #385, opened 2026-05-16
fix(api): call auth handlers in stream and wait run endpoints

arbel03:
## Summary

- The `create_and_stream_run` (`POST /threads/{id}/runs/stream`) and `wait_for_run` (`POST /threads/{id}/runs/wait`) endpoints were missing the `handle_event` call for the `threads.create_run` auth action.
- This meant `@auth.on.threads.create_run` handlers were only invoked for the non-streaming `POST /threads/{id}/runs` endpoint, but silently skipped for the streaming and wait variants.
- This broke any authorization logic registered via `@auth.on.threads.create_run` (e.g. usage tracking, rate limiting, metadata injection) when using the streaming endpoint.

## Changes

Added the same `build_auth_context` + `handle_event` + config/context merge pattern (already present in `create_run`) to both `create_and_stream_run` and `wait_for_run`.

## Test plan

- [x] Verify `@auth.on.threads.create_run` handler is called when using `POST /threads/{id}/runs/stream`
- [x] Verify `@auth.on.threads.create_run` handler is called when using `POST /threads/{id}/runs/wait`
- [x] Verify handler can reject requests (returns False or raises HTTPException)
- [x] Verify handler can modify config/context via return dict
- [x] Verify existing `POST /threads/{id}/runs` behavior is unchanged

## Summary by CodeRabbit

* **New Features**
  * Run management (create, streaming, and wait) now honors `config`/`context` overrides returned by authorization handlers during run initialization, including correct precedence rules.
* **Bug Fixes**
  * Authorization denials (403) now consistently halt run preparation before it proceeds.
  * Empty handler returns no longer trigger unintended fallback behavior for `config`/`context`.
* **Tests**
  * Added unit coverage for handler dispatch, denial behavior, and `config`/`context` merging across streaming and wait flows.
* **Chores**
  * Version bumps for both the API and CLI (0.9.24 → 0.9.25).

<h3>Greptile Summary</h3>

The PR consistently applies create-run authorization across all threaded run creation variants.
- Extracts the authorization and config/context override logic into a shared helper.
- Calls the helper from standard, streaming, and wait run endpoints.
- Adds regression coverage for handler invocation, denial, and override semantics.
- Keeps API and CLI package versions synchronized at 0.9.25.

The PR appears safe to merge.

No blocking failure remains; the previous missing-test and duplicated-auth-logic findings are addressed by endpoint regression coverage and the shared authorization helper.

| Filename | Overview |
|----------|----------|
| libs/aegra-api/src/aegra_api/api/runs.py | Centralizes create-run authorization and applies it before preparation in all three run creation endpoints. |
| libs/aegra-api/tests/unit/test_api/test_runs.py | Covers the shared helper's returned-filter and in-place mutation semantics. |
| libs/aegra-api/tests/unit/test_api/test_runs_streaming.py | Adds streaming endpoint regression tests for authorization invocation, denial, and request overrides. |
| libs/aegra-api/tests/unit/test_api/test_runs_wait.py | Adds wait endpoint regression tests for authorization invocation, denial, and request overrides. |
| libs/aegra-api/pyproject.toml | Bumps the API package version to 0.9.25. |
| libs/aegra-cli/pyproject.toml | Keeps the CLI package version synchronized at 0.9.25. |
| uv.lock | Synchronizes workspace package metadata with the new API and CLI versions. |

</details>

```mermaid
sequenceDiagram
    participant Client
    participant RunsAPI
    participant AuthHandler
    participant RunPreparation

    Client->>RunsAPI: Create, stream, or wait for run
    RunsAPI->>AuthHandler: threads.create_run
    alt Request denied
        AuthHandler-->>Client: 403 Forbidden
    else Request allowed
        AuthHandler-->>RunsAPI: Config/context overrides
        RunsAPI->>RunPreparation: Prepare merged request
        RunPreparation-->>Client: Run response or stream
    end
```
</details>

<sub>Reviews (4): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/stream-endp..."](https://github.com/aegra/aegra/commit/460d10c20b6fa0a8a2dca61158ec038f04e8563f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32440260)</sub>

**Context used:**

- Knowledge Base — [Authentication and Authorization](https://app.greptile.com/aegra/-/custom-context/knowledge-base/aegra/aegra/-/docs/auth.md)

### DIFF
--- a/libs/aegra-api/pyproject.toml
+++ b/libs/aegra-api/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "aegra-api"
-version = "0.9.24"
+version = "0.9.25"
 description = "Aegra core API - Self-hosted Agent Protocol server"
 readme = "README.md"
 requires-python = ">=3.12"
--- a/libs/aegra-api/src/aegra_api/api/runs.py
+++ b/libs/aegra-api/src/aegra_api/api/runs.py
@@ -41,6 +41,24 @@ logger = structlog.getLogger(__name__)
 DEFAULT_STREAM_MODES = ["values"]
 
 
+async def _apply_create_run_auth(user: User, thread_id: str, request: RunCreate) -> None:
+    """Authorize threads.create_run and merge config/context overrides into request.
+
+    Handler-returned filter dict wins; otherwise fall back to in-place value mutations.
+    """
+    ctx = build_auth_context(user, "threads", "create_run")
+    value = {**request.model_dump(), "thread_id": thread_id}
+    filters = await handle_event(ctx, value)
+
+    source = filters if filters is not None else value
+    config_overrides = source.get("config")
+    if isinstance(config_overrides, dict):
+        request.config = {**(request.config or {}), **config_overrides}
+    context_overrides = source.get("context")
+    if isinstance(context_overrides, dict):
+        request.context = {**(request.context or {}), **context_overrides}
+
+
 @router.post("/threads/{thread_id}/runs", response_model=Run, responses={**NOT_FOUND, **CONFLICT})
 async def create_run(
     thread_id: str,
@@ -59,25 +77,7 @@ async def create_run(
     if existing_thread and existing_thread.user_id != user.identity:
         raise HTTPException(404, f"Thread '{thread_id}' not found")
 
-    # Authorization check (create_run action on threads resource)
-    ctx = build_auth_context(user, "threads", "create_run")
-    value = {**request.model_dump(), "thread_id": thread_id}
-    filters = await handle_event(ctx, value)
-
-    # If handler modified config/context, update request
-    if filters:
-        if "config" in filters and isinstance(filters["config"], dict):
-            request.config = {**(request.config or {}), **filters["config"]}
-        if "context" in filters and isinstance(filters["context"], dict):
-            request.context = {**(request.context or {}), **filters["context"]}
-    else:
-        value_config = value.get("config")
-        if isinstance(value_config, dict):
-            request.config = {**(request.config or {}), **value_config}
-
-        value_context = value.get("context")
-        if isinstance(value_context, dict):
-            request.context = {**(request.context or {}), **value_context}
+    await _apply_create_run_auth(user, thread_id, request)
 
     _run_id, run, _job = await _prepare_run(session, thread_id, request, user, initial_status="pending")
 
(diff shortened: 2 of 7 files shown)


ibbybuilds (changes requested):
thanks @arbel03, this is a real gap and you diagnosed it right. `@auth.on.threads.create_run` handlers were being skipped on the stream and wait endpoints, so anyone blocked on `POST /runs` could still start the same run through `/runs/stream` or `/runs/wait`. the fix is in the right place too, handler call after the ownership 404 check and before `_prepare_run`, matching `create_run` exactly.

two things before this can go in:

**tests are required.** this is a security fix with zero tests, and the project rule is bug fixes need a regression test, no exceptions. a silent authz bypass with no test can silently come back. please add cases to `test_runs_streaming.py` and `test_runs_wait.py`: handler returns a deny -> 403, handler returns a config/context dict -> merged into the run, happy path unchanged. that also locks the behavior so a future refactor can't quietly drop the handler again.

**rebase.** it's conflicting against main right now.

one small thing while you're in there: the auth block is now copy pasted across three endpoints. worth pulling into a small `_apply_create_run_auth(user, thread_id, request)` helper so the three run-creation paths can't drift apart again. not blocking.

logic is ready to ship. add the tests and rebase and i'll merge this quickly, it's the highest priority open PR right now.

arbel03:
@ibbybuilds

ibbybuilds (approved):
approving. i pulled this down and verified it properly rather than reading the diff.

the gap is real and i confirmed it against main: create_run dispatches the threads.create_run handler, create_and_stream_run and wait_for_run do not. so any deployment using a handler for deny logic, rate limiting or usage metering was bypassable by calling /runs/stream or /runs/wait instead of /runs. the fix puts the call in the right place on both, after the ownership 404 and before _prepare_run, matching create_run exactly.

the extraction into _apply_create_run_auth is the right call. three copies of that block would have drifted, and a future endpoint that forgets it reintroduces the same hole.

the `filters is not None` change is a real catch i had missed. handle_event documents dict and None as distinct signals, so a handler returning {} means allowed with no filters. the old `if filters:` treated that as falsy and fell through to applying in-place value mutations, which is the opposite of what the handler asked for. good find.

on the tests, i checked they actually fail without the fix rather than just passing alongside it. removing a single call site fails 6 of them. that is what i wanted, a test that pins the behavior so this cannot silently come back.

verified locally on your branch:
- 36 tests in the three run test files pass
- full unit suite 1403 passed
- 157 auth tests pass, no regressions on the existing create_run path

one follow-up, not blocking this: the same three-endpoint pattern means anything else added to create_run in future needs the same treatment. worth a note in the auth docs at some point, i will handle that.

thanks for adding the regression tests, and sorry this sat as long as it did.

ibbybuilds (approved):
LGTM

PR #422, opened 2026-06-11
fix: replace config btree index with md5 expression index to support large assistant configs

### LINKED ISSUE #421: idx_assistant_user_graph_config btree index fails for large config payloads — agents with long system prompts cannot be saved
### Summary

Any `POST /assistants` (create) or `PATCH /assistants/{id}` (update) where `config.configurable` contains a large value — most commonly a long system prompt stored under `config.configurable.system_prompt` — fails with a PostgreSQL error:

```
index row size NNNN exceeds btree version 4 maximum 2704 for index
"idx_assistant_user_graph_config"
```

Aegra has no handler for this exception, so it surfaces as a 500 and the client receives no actionable error message.

---

### Root cause

Revision `76dfdbe90d2b` (`add_version_table`) creates:

```sql
CREATE UNIQUE INDEX idx_assistant_user_graph_config
    ON assistant (user_id, graph_id, config);
```

PostgreSQL btree index entries are capped at roughly **2 704 bytes** (one third of an 8 KB page). The `config` column is a JSONB object whose `configurable` sub-key holds agent configuration — including a system prompt that, in practice, routinely exceeds 2 704 bytes. Every write to an assistant whose serialised `config` is larger than that limit hits the cap.

---

### Proposed fix

Replace the verbatim-column btree index with a fixed-width **expression index** over `md5(config::text)`:

```sql
CREATE UNIQUE INDEX idx_assistant_user_graph_config
    ON assistant (user_id, graph_id, md5(config::text));
```

`md5(config::text)` is always 32 bytes, so the index entry never exceeds the btree limit regardless of prompt size. The uniqueness guarantee is preserved: PostgreSQL stores JSONB in canonical form (normalised key order, no insignificant whitespace), so `config::text` is deterministic for logically-equal configs. `md5` is a PostgreSQL built-in — no `pgcrypto` extension is required.

**New Alembic migration:**

```python
def upgrade() -> None:
    op.drop_index("idx_assistant_user_graph_config", table_name="assistant")
    op.execute(
        "CREATE UNIQUE INDEX idx_assistant_user_graph_config"
        " ON assistant (user_id, graph_id, md5(config::text))"
    )

def downgrade() -> None:
    # WARNING: will fail if any row's config::text already exceeds ~2 704 bytes.
    op.drop_index("idx_assistant_user_graph_config", table_name="assistant")
    op.create_index(
        "idx_assistant_user_graph_config",
        "assistant",
        ["user_id", "graph_id", "config"],
        unique=True,
    )
```

**`orm.py` `__table_args__` update** (keeps the ORM declaration in sync with the migration):

```python
# before
Index("idx_assistant_user_graph_config", "user_id", "graph_id", "config", unique=True),

# after
Index("idx_assistant_user_graph_config", "user_id", "graph_id", text("md5(config::text)"), unique=True),
```

---

### Notes

- The `assistant_service.py` deduplication query (`AssistantORM.config == config`) does not need to change — it uses JSONB equality on the column directly, not the index expression.
- `DROP + CREATE` are intentionally **not** `CONCURRENTLY` so they run inside a transaction: if `CREATE` fails, `DROP` rolls back and the original index survives.
- The downgrade is **forward-only safe**: re-creating the original btree index on a database that already contains large-config rows will fail. This should be documented in the migration.

---

### Environment

- `aegra-api` version: `0.9.18`
- PostgreSQL: 14+ (reproduced on 14, 15, 16)
- Triggered by: saving an assistant variant where `len(json.dumps(config)) > ~2700`

prithvirao13:
Fixes #421

## Problem

Revision `76dfdbe90d2b` created a UNIQUE btree index over `(user_id, graph_id, config)` on the `assistant` table. PostgreSQL btree index entries are capped at ~2 704 bytes (1/3 of an 8 KB page). The `config` column is a JSONB object that holds agent configuration under `config.configurable` — including a system prompt that routinely exceeds 2 704 bytes. Every `INSERT` or `UPDATE` whose serialised `config` is larger than that limit fails with:

```
index row size NNNN exceeds btree version 4 maximum 2704 for index
"idx_assistant_user_graph_config"
```

Aegra has no handler for this error, so it surfaces as a 500 with no actionable message.

## Fix

Replace the verbatim-column btree with a fixed-width **expression index** over `md5(config::text)`:

```sql
CREATE UNIQUE INDEX idx_assistant_user_graph_config
    ON assistant (user_id, graph_id, md5(config::text));
```

`md5(config::text)` is always 32 bytes, so the index entry is never larger than the btree limit regardless of config size. The uniqueness guarantee is preserved: PostgreSQL serialises JSONB in canonical form, making `config::text` deterministic for logically-equal configs. `md5` is a PostgreSQL built-in — no `pgcrypto` extension is required.

## Changes

| File | Change |
|---|---|
| `libs/aegra-api/alembic/versions/20260612000000_hash_assistant_config_index.py` | New migration: drop old btree, create md5 expression index |
| `libs/aegra-api/src/aegra_api/core/orm.py` | Update `__table_args__` to match the new expression index |

No changes to `assistant_service.py` — the app-level deduplication query (`AssistantORM.config == config`) uses JSONB equality on the column directly and remains correct.

## Downgrade note

The downgrade path re-creates the original verbatim-config btree. This will fail on any database that already contains a row whose `config::text` exceeds ~2 704 bytes. This migration should be treated as **forward-only** on production databases that have stored large-config assistants.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved assistant configuration persistence by changing the uniqueness index to use a fixed-size hash derived from the configuration content, preventing PostgreSQL btree index entry-size failures with very large prompts.
  * Included the matching database migration for the updated index strategy; rollback may fail if existing rows contain exceptionally large configuration text values.
* **Tests**
  * Added an integration regression test that inserts an assistant with a large, deterministic configuration and verifies it’s persisted correctly.

<h3>Greptile Summary</h3>

This PR fixes a PostgreSQL btree size limit error that caused any `INSERT` or `UPDATE` on the `assistant` table to fail with a 500 when the serialized `config` JSONB exceeded ~2 704 bytes, which is the btree entry cap. The fix replaces the verbatim-column index with a fixed-width `md5(config::text)` expression index — the 32-byte hash is always within the limit while still enforcing uniqueness, since PostgreSQL stores JSONB in canonical form making the cast deterministic.

- **Migration** (`20260612000000_hash_assistant_config_index.py`): drops the old btree and creates the md5 expression index; `revision = "b88bb61be638"` is unique and `down_revision = "c7d1f2a4b6e8"` correctly chains to `cron_cascade_and_claim`, the current HEAD — the duplicate-revision issue from the previous review iteration has been resolved.
- **ORM** (`orm.py`): `__table_args__` is updated to use `text("md5(config::text)")` so the schema produced by `Base.metadata.create_all()` matches the migration.
- **Test** (`test_assistant_large_config_db.py`): new integration regression test inserts a >2 704-byte config, asserts the row was persisted correctly, and skips gracefully when no live database is available.

Safe to merge — the migration chain is valid, the ORM and migration are consistent, and the fix is well-scoped to the reported btree overflow.

The migration revision is unique, its down_revision correctly points to the current HEAD, the ORM expression index matches the migration SQL, and the application-level deduplication query (config == config) is unaffected. The md5 approach is a standard, well-understood pattern for this class of problem in PostgreSQL.

| Filename | Overview |
|----------|----------|
| libs/aegra-api/alembic/versions/20260612000000_hash_assistant_config_index.py | New Alembic migration (revision b88bb61be638) that drops the verbatim-config btree index and creates an md5 expression index; revision ID is unique and down_revision correctly chains to cron_cascade_and_claim (c7d1f2a4b6e8), the current HEAD. |
| libs/aegra-api/src/aegra_api/core/orm.py | Updates the Assistant ORM __table_args__ to use text("md5(config::text)") in the expression index, keeping it in sync with the Alembic migration. |
| libs/aegra-api/tests/integration/test_assistant_large_config_db.py | New integration regression test that inserts an assistant with a >2704-byte config, verifies the save succeeded, and cleans up; correctly skips when no live database is available. |

</details>

<h3>Sequence Diagram</h3>

<a href="#gh-light-mode-only">

```mermaid
sequenceDiagram
    participant App as Application
    participant PG as PostgreSQL

    Note over App,PG: Before fix — large config fails
    App->>PG: "INSERT INTO assistant (user_id, graph_id, config=large_jsonb)"
    PG-->>App: ERROR: index row size 3500 exceeds btree max 2704

    Note over App,PG: After fix — md5 expression index
    App->>PG: "INSERT INTO assistant (user_id, graph_id, config=large_jsonb)"
    PG->>PG: Index check: md5(config::text) → 32-byte hash
    PG->>PG: Unique check on (user_id, graph_id, hash)
    PG-->>App: OK (row inserted)

    Note over App,PG: Deduplication query unchanged
    App->>PG: "SELECT * WHERE config = :config (JSONB equality)"
    PG-->>App: existing row (if duplicate)
```

</a>
<a href="#gh-dark-mode-only">

```mermaid
%%{init: {'theme': 'base', 'themeVariables': {"darkMode": true, "background": "#0d1117", "primaryColor": "#21262d", "primaryTextColor": "#e6edf3", "primaryBorderColor": "#8b949e", "lineColor": "#8b949e", "textColor": "#e6edf3", "edgeLabelBackground": "#161b22", "actorBkg": "#21262d", "actorBorder": "#8b949e", "actorTextColor": "#e6edf3", "actorLineColor": "#8b949e", "signalColor": "#8b949e", "signalTextColor": "#e6edf3", "noteBkgColor": "#373320", "noteBorderColor": "#d4a72c", "noteTextColor": "#f0e6c0", "labelBoxBkgColor": "#21262d", "labelBoxBorderColor": "#8b949e", "labelTextColor": "#e6edf3", "loopTextColor": "#e6edf3", "activationBkgColor": "#30363d", "activationBorderColor": "#8b949e"}}}%%
sequenceDiagram
    participant App as Application
    participant PG as PostgreSQL

    Note over App,PG: Before fix — large config fails
    App->>PG: "INSERT INTO assistant (user_id, graph_id, config=large_jsonb)"
    PG-->>App: ERROR: index row size 3500 exceeds btree max 2704

    Note over App,PG: After fix — md5 expression index
    App->>PG: "INSERT INTO assistant (user_id, graph_id, config=large_jsonb)"
    PG->>PG: Index check: md5(config::text) → 32-byte hash
    PG->>PG: Unique check on (user_id, graph_id, hash)
    PG-->>App: OK (row inserted)

    Note over App,PG: Deduplication query unchanged
    App->>PG: "SELECT * WHERE config = :config (JSONB equality)"
    PG-->>App: existing row (if duplicate)
```

</a>

<sub>Reviews (4): Last reviewed commit: ["fix: initialise assistant\_table before t..."](https://github.com/aegra/aegra/commit/992967d980c08bf6613147996baf3268ad063eda) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36893774)</sub>

### DIFF
--- a/libs/aegra-api/alembic/versions/20260612000000_hash_assistant_config_index.py
+++ b/libs/aegra-api/alembic/versions/20260612000000_hash_assistant_config_index.py
@@ -36,15 +36,15 @@ database that already contains a row whose ``config::text`` exceeds
 migration as **forward-only** on production databases that have been used
 with large agent configs.
 
-Revision ID: a1b2c3d4e5f6
-Revises: f1a2b3c4d5e6
+Revision ID: b88bb61be638
+Revises: c7d1f2a4b6e8
 Create Date: 2026-06-12 00:00:00.000000
 """
 
 from alembic import op
 
-revision = "a1b2c3d4e5f6"
-down_revision = "f1a2b3c4d5e6"
+revision = "b88bb61be638"
+down_revision = "c7d1f2a4b6e8"
 branch_labels = None
 depends_on = None
 
--- a/libs/aegra-api/alembic/versions/20260612000000_hash_assistant_config_index.py
+++ b/libs/aegra-api/alembic/versions/20260612000000_hash_assistant_config_index.py
@@ -1,0 +1,74 @@
+"""Replace verbatim-config btree index with md5 expression index.
+
+The unique btree index ``idx_assistant_user_graph_config`` was introduced in
+revision ``76dfdbe90d2b`` (add_version_table) as::
+
+    CREATE UNIQUE INDEX idx_assistant_user_graph_config
+        ON assistant (user_id, graph_id, config);
+
+PostgreSQL btree index entries are capped at roughly **2 704 bytes** (one
+third of an 8 KB page). The ``config`` column is a JSONB object whose
+``configurable`` sub-key holds agent configuration — including a system
+prompt that routinely exceeds 2 704 bytes. Every ``INSERT`` or ``UPDATE``
+whose serialised ``config`` is larger than that limit fails with::
+
+    index row size NNNN exceeds btree version 4 maximum 2704 for index
+    "idx_assistant_user_graph_config"
+
+This migration replaces the verbatim-column btree with a fixed-width
+**expression index** over ``md5(config::text)``::
+
+    CREATE UNIQUE INDEX idx_assistant_user_graph_config
+        ON assistant (user_id, graph_id, md5(config::text));
+
+``md5(config::text)`` is always 32 bytes, so the index entry never exceeds
+the btree limit regardless of payload size.  The uniqueness guarantee is
+preserved: PostgreSQL stores JSONB in canonical form (normalised key order,
+no insignificant whitespace), so ``config::text`` is deterministic for
+logically-equal configs.  ``md5`` is a PostgreSQL built-in — no
+``pgcrypto`` extension is required.
+
+Downgrade note
+--------------
+The downgrade re-creates the original btree index.  This will fail on any
+database that already contains a row whose ``config::text`` exceeds
+~2 704 bytes (i.e. the exact problem this migration fixes).  Treat this
+migration as **forward-only** on production databases that have been used
+with large agent configs.
+
+Revision ID: a1b2c3d4e5f6
+Revises: f1a2b3c4d5e6
+Create Date: 2026-06-12 00:00:00.000000
+"""
+
+from alembic import op
+
+revision = "a1b2c3d4e5f6"
+down_revision = "f1a2b3c4d5e6"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.drop_index("idx_assistant_user_graph_config", table_name="assistant")
+    # CREATE INDEX cannot use CONCURRENTLY inside a transaction, and Alembic
+    # migrations run inside a transaction by default.  The brief ShareLock on
+    # the assistant table during the index build is acceptable: the table is
+    # small and this migration runs once at deployment time.
+    op.execute(
+        "CREATE UNIQUE INDEX idx_assistant_user_graph_config"
+        " ON assistant (user_id, graph_id, md5(config::text))"
+    )
+
+
+def downgrade() -> None:
+    # WARNING: re-creating the verbatim-config btree will fail if any row's
+    # config::text already exceeds ~2 704 bytes.  Only safe on databases that
+    # have never stored a large-config assistant.
+    op.drop_index("idx_assistant_user_graph_config", table_name="assistant")
+    op.create_index(
+        "idx_assistant_user_graph_config",
+        "assistant",
+        ["user_id", "graph_id", "config"],
+        unique=True,
+    )
--- a/libs/aegra-api/alembic/versions/20260612000000_hash_assistant_config_index.py
+++ b/libs/aegra-api/alembic/versions/20260612000000_hash_assistant_config_index.py
@@ -56,8 +56,7 @@ def upgrade() -> None:
     # the assistant table during the index build is acceptable: the table is
     # small and this migration runs once at deployment time.
     op.execute(
-        "CREATE UNIQUE INDEX idx_assistant_user_graph_config"
-        " ON assistant (user_id, graph_id, md5(config::text))"
+        "CREATE UNIQUE INDEX idx_assistant_user_graph_config ON assistant (user_id, graph_id, md5(config::text))"
     )
 
 
--- a/libs/aegra-api/src/aegra_api/core/orm.py
+++ b/libs/aegra-api/src/aegra_api/core/orm.py
@@ -114,7 +114,7 @@ class Assistant(Base):
             "idx_assistant_user_graph_config",
             "user_id",
             "graph_id",
-            "config",
+            text("md5(config::text)"),
             unique=True,
         ),
     )
--- a/libs/aegra-api/tests/integration/test_assistant_large_config_db.py
+++ b/libs/aegra-api/tests/integration/test_assistant_large_config_db.py
@@ -27,6 +27,7 @@ async def test_insert_assistant_with_large_configurable_prompt_succeeds() -> Non
     engine = create_async_engine(settings.db.database_url)
 
     try:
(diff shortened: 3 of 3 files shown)


ibbybuilds:
@
@prithvirao13 lint is failing on ruff format. run `uv run ruff format .` and push, one line change to the CREATE INDEX string in the migration.

also needs a regression test before merge, our rule for bug fixes. insert an assistant whose config serializes past 2704 bytes (big system prompt under config.configurable) and assert the create succeeds. it fails on main today and passes with your fix, which is the point. integration level against the db is fine.

last thing, can you confirm down_revision c7d1f2a4b6e8 is still the alembic head? main moved since you opened this, want to be sure the chain stays linear on a fresh db.
@

prithvirao13:
Addressed the review items:

- Ran Ruff formatting; the CREATE INDEX string is now formatted as expected.
- Added a DB regression test: `libs/aegra-api/tests/integration/test_assistant_large_config_db.py`. It inserts an assistant with a large `config.configurable.system_prompt` (> btree tuple limit) and asserts the insert succeeds.
- Confirmed upstream `main` still has Alembic head `c7d1f2a4b6e8`, so this migration's `down_revision` remains correct and the chain stays linear.

Local verification:
- `uvx ruff format --check .` ✅
- `uvx ruff check .` ✅
- `uv run --package aegra-api pytest libs/aegra-api/tests/unit libs/aegra-api/tests/integration -q` ✅ (1550 passed, 1 skipped locally because no default Postgres)
- Started a disposable PostgreSQL 18 container, ran `alembic upgrade head`, then ran the new regression test against it: ✅ passed.

ibbybuilds:
@prithvirao13 codeql is failing on the new test, not the migration. test_assistant_large_config_db.py:37, assistant_table can be read before assignment: it is only bound inside the inner try, and codeql cannot prove the except branch (pytest.skip) always exits, so the `if assistant_table is None` on line 37 is flagged as use-before-init.

init it before the try:

    assistant_table = None
    try:
        async with engine.begin() as conn:
            ...
            assistant_table = await conn.scalar(...)
    except (OperationalError, OSError) as exc:
        pytest.skip(...)

that clears the alert. everything else is green now, this is the last thing.

PR #448, opened 2026-07-06
fix(api): forward run-level interrupts to graphs

### LINKED ISSUE #442: Run-level interrupt_before / interrupt_after are silently ignored — never reach graph.astream
## Summary

Setting `interrupt_before` (or `interrupt_after`) on a run — via `POST /threads/{id}/runs/wait`, `/runs`, or crons — has no effect. The run executes to completion as if the field were absent. The value is stored on the LangGraph run `config` dict but never passed to `graph.astream` / `graph.astream_events`, which read those options only from their keyword arguments, not from `config`.

This breaks the documented run-level interrupt feature (`docs/guides/human-in-the-loop.mdx`: "set interrupt points without modifying the graph code, using the `interrupt_before` and `interrupt_after` parameters on the run").

Observed against `aegra-api` 0.9.23 (latest at time of writing; identical on 0.9.16). Reproduced both in a standalone LangGraph script and on a live deployment.

## Root cause

Two places, one bug:

1. `libs/aegra-api/src/aegra_api/services/run_executor.py` — `_build_run_config()` writes the value **into the config dict**:

```python
if job.behavior.interrupt_before is not None:
    items = job.behavior.interrupt_before
    config["interrupt_before"] = items if isinstance(items, list) else [items]
if job.behavior.interrupt_after is not None:
    items = job.behavior.interrupt_after
    config["interrupt_after"] = items if isinstance(items, list) else [items]
```

2. `libs/aegra-api/src/aegra_api/services/graph_streaming.py` — `stream_graph_events()` then calls the graph **without forwarding those as kwargs**:

```python
graph.astream_events(
    input_data,
    config,
    context=context,
    version="v2",
    stream_mode=list(stream_modes_set),
    subgraphs=subgraphs,
)
# ...and the astream path:
graph.astream(
    input_data,
    config,
    context=context,
    stream_mode=list(stream_modes_set),
    output_keys=output_keys,
    subgraphs=subgraphs,
)
```

`interrupt_before` / `interrupt_after` are **keyword parameters of `Pregel.astream` / `astream_events`** in LangGraph. They are not read from `RunnableConfig`. So putting them in `config` is a no-op, and the run proceeds straight through every node, including the model.

## Secondary issue: `"*"` ("all nodes") sentinel is coerced into `["*"]`

Even once the value is forwarded, `_build_run_config` wraps a bare string in a list:

```python
config["interrupt_before"] = items if isinstance(items, list) else [items]
```

LangGraph's "interrupt before all nodes" sentinel is the bare string `"*"` (`langgraph.types.All`). Coercing it to `["*"]` makes LangGraph interpret it as "interrupt before a node literally named `*`", which matches nothing — so `interrupt_before: "*"` would still run to completion even after fix (1). The `"*"` case needs to be preserved as the bare sentinel rather than list-wrapped.

## Reproduction

Via the API:

1. Any graph with at least one node.
2. `POST /threads/{id}/runs/wait` with body `{"assistant_id": "...", "input": {...}, "interrupt_before": "*"}` (or a named node list like `["tools"]`).
3. Expected: the run halts before the first node; thread left `interrupted` with a non-empty `next`.
4. Actual: the run executes to completion; no interrupt.

### Standalone reproduction (isolates the config→kwarg handoff)

Self-contained — only needs `langgraph`. No server, DB, or agent graph. A 2-node graph (`a` → `b`) compiled without compile-time interrupts, exactly as Aegra compiles user graphs:

```python
import asyncio
import uuid

from langgraph.checkpoint.memory import MemorySaver
from langgraph.graph import END, START, StateGraph
from typing_extensions import TypedDict


class State(TypedDict):
    n: int


def build():
    b = StateGraph(State)
    b.add_node("a", lambda s: {"n": s["n"] + 1})
    b.add_node("b", lambda s: {"n": s["n"] + 10})
    b.add_edge(START, "a")
    b.add_edge("a", "b")
    b.add_edge("b", END)
    # Compile WITHOUT compile-time interrupts, exactly like Aegra compiles user graphs.
    return b.compile(checkpointer=MemorySaver())


async def run(label, *, config_interrupt=None, kwarg_interrupt=None):
    g = build()
    cfg = {"configurable": {"thread_id": str(uuid.uuid4())}}
    if config_interrupt is not None:
        cfg["interrupt_before"] = config_interrupt  # where Aegra writes it
    kwargs = {}
    if kwarg_interrupt is not None:
        kwargs["interrupt_before"] = kwarg_interrupt
    async for _ in g.astream({"n": 0}, cfg, **kwargs):
        pass
    st = g.get_state(cfg)
    print(f"{label}: next={st.next!r} -> {'HALTED' if st.next else 'RAN TO COMPLETION'}")


async def main():
    # A: interrupt_before in the config dict, no kwarg  <- what Aegra does
    await run("A  config interrupt_before=['a'], no kwarg (Aegra path)", config_interrupt=["a"])
    # B: interrupt_before passed as the astream kwarg   <- positive control
    await run("B  kwarg  interrupt_before=['a']        (control)", kwarg_interrupt=["a"])
    # C: interrupt_before=["*"]  (Aegra's list-coercion of "*")
    await run("C  kwarg  interrupt_before=['*']        (Aegra '*' coercion)", kwarg_interrupt=["*"])
    # C': interrupt_before="*"  (the bare All sentinel)
    await run("C' kwarg  interrupt_before='*'          (correct sentinel)", kwarg_interrupt="*")


asyncio.run(main())
```

Output (aegra-api 0.9.23):

```
A  config interrupt_before=['a'], no kwarg (Aegra path): next=() -> RAN TO COMPLETION
B  kwarg  interrupt_before=['a']        (control): next=('a',) -> HALTED
C  kwarg  interrupt_before=['*']        (Aegra '*' coercion): next=() -> RAN TO COMPLETION
C' kwarg  interrupt_before='*'          (correct sentinel): next=('a',) -> HALTED
```

**A** is exactly Aegra's behavior and runs straight through. **B** is the positive control: the same value passed as the `astream` keyword argument halts the run — so LangGraph and the graph are fine; only the config→kwarg handoff is missing. **C** vs **C'** isolates the secondary `"*"` coercion bug independently.

## Suggested fix

In `stream_graph_events` (or where the run config is assembled for streaming), lift `interrupt_before` / `interrupt_after` out of `config` and pass them as keyword arguments to `graph.astream` / `graph.astream_events`, and preserve the `"*"` sentinel instead of wrapping it in a list. Roughly:

```python
interrupt_before = config.pop("interrupt_before", None)
interrupt_after = config.pop("interrupt_after", None)
# normalize ["*"] -> "*"
graph.astream(
    input_data,
    config,
    ...,
    interrupt_before=interrupt_before,
    interrupt_after=interrupt_after,
)
```

Daniyal0100101:
## Description

This fixes run-level interrupt handling for API and cron-triggered runs. `interrupt_before` and `interrupt_after` are currently stored in the run config, but `stream_graph_events` did not forward them to LangGraph's `graph.astream` / `graph.astream_events` calls, so they were silently ignored during execution.

The fix extracts those interrupt settings from a copy of the config and forwards them as explicit kwargs to both streaming paths. It also preserves LangGraph's all-nodes interrupt sentinel: when the run config contains `["*"]`, it is normalized back to the bare string `"*"`, which is what LangGraph expects.

## Type of Change

- [x] `fix`: Bug fix
- [x] `test`: Tests added/updated

## Related Issues

Fixes #442

## Changes Made

- Added `_extract_interrupt_kwargs` to lift `interrupt_before` / `interrupt_after` from `RunnableConfig` before calling LangGraph streaming methods.
- Added `_normalize_interrupt_value` to preserve the `"*"` all-nodes sentinel correctly.
- Forwarded interrupt kwargs to both `graph.astream` and `graph.astream_events`.
- Added regression coverage for normal interrupt lists and the all-nodes sentinel across both streaming branches.
- Bumped `aegra-api` and `aegra-cli` from `0.9.24` to `0.9.25`, with matching lockfile and OpenAPI metadata updates.

## Testing

- [ ] Integration tests added/updated — not needed; this is isolated stream forwarding behavior
- [ ] E2E tests added/updated — not needed for this narrow service-level bug fix

Validation run locally:

```bash
uv run ruff check .
uv run ruff format --check .
uv run --package aegra-api pytest libs/aegra-api/tests/unit/test_services/test_graph_streaming_events_mode.py -q
uv run --package aegra-api pytest libs/aegra-api/tests/unit/test_services -q
git diff --check
```

Results:

- `ruff check`: passed
- `ruff format --check`: passed
- targeted stream test file: 6 passed, 1 warning
- service unit tests: 698 passed, 1 warning
- `git diff --check`: passed

Notes on broader checks:

- `make type-check` currently reports pre-existing diagnostics outside this change, including existing diagnostics in `graph_streaming.py`; I did not broaden this PR to type-clean unrelated code.
- `make test` starts E2E tests and timed out in this local environment because the E2E/server setup is not running here. I validated the affected service layer directly instead.
- `uv run --package aegra-api pytest libs/aegra-api/tests/ --ignore=libs/aegra-api/tests/e2e -q` currently has an unrelated existing failure in `test_langgraph_service.py::TestLangGraphServiceConfig::test_initialize_env_var_override`; this same test passes on a clean detached `origin/main` worktree in this environment, so I treated it as environment/worktree-specific rather than caused by this PR.

## Checklist

- [x] Linting checks pass (`uv run ruff check .`)
- [ ] Type checking passes (`make type-check`) — existing project diagnostics remain outside this change
- [x] Relevant tests pass
- [x] Documentation updated (not user-facing behavior docs; release/OpenAPI metadata updated for version bump)
- [x] Commit messages follow Conventional Commits

## Summary by CodeRabbit

* **New Features**
  * Streaming now respects interrupt settings, forwarding `interrupt_before`/`interrupt_after` correctly for both value and event streaming modes, including wildcard (`"*"`) handling.

* **Chores**
  * Bumped the release version to **0.9.25** across the API package, CLI package, and OpenAPI documentation.

* **Tests**
  * Added async unit tests to verify interrupt forwarding behavior and wildcard preservation for both streaming modes.

<h3>Greptile Summary</h3>

This PR fixes a silent bug where `interrupt_before` / `interrupt_after` stored in the run's `RunnableConfig` were never forwarded to LangGraph's `graph.astream` or `graph.astream_events` calls, meaning run-level interrupts were silently ignored for API and cron-triggered runs.

- Adds `_extract_interrupt_kwargs` to peel the two interrupt keys from a shallow copy of the config, and `_normalize_interrupt_value` to convert the `[\"*\"]` list form back to the bare `\"*\"` string sentinel LangGraph expects; both are then spread into `**interrupt_kwargs` on the two streaming call sites.
- Adds four regression tests using a new `_RecordingGraph` async-generator mock, covering normal interrupt lists and the wildcard sentinel for both the `astream` and `astream_events` code paths.

Safe to merge — the change is a narrow, well-scoped forwarding fix with no mutations to shared state and no new failure modes introduced.

The helper functions are pure and side-effect-free: _extract_interrupt_kwargs operates on a shallow copy so the original config is never mutated, and _normalize_interrupt_value is a simple value transformation. Both streaming call sites now receive the interrupt kwargs consistently. All four regression tests cover normal lists and the wildcard sentinel for both astream and astream_events, closing the coverage gap that existed in the previous iteration.

**Files Needing Attention:** No files require special attention.

| Filename | Overview |
|----------|----------|
| libs/aegra-api/src/aegra_api/services/graph_streaming.py | Adds _extract_interrupt_kwargs / _normalize_interrupt_value helpers and threads the extracted kwargs into both astream and astream_events call sites; logic is correct and matches the stated LangGraph contract. |
| libs/aegra-api/tests/unit/test_services/test_graph_streaming_events_mode.py | Adds _RecordingGraph mock and four tests covering interrupt forwarding and ["*"]->"*" normalization for both streaming branches; addresses the previously noted gap in sentinel test coverage for astream_events. |

</details>

```mermaid
sequenceDiagram
    participant Caller as API / Cron caller
    participant SGE as stream_graph_events
    participant EIK as _extract_interrupt_kwargs
    participant NIV as _normalize_interrupt_value
    participant LG as LangGraph graph

    Caller->>SGE: "config{interrupt_before, interrupt_after, ...}"
    SGE->>EIK: config (with interrupt keys)
    EIK->>NIV: "value (e.g. ["*"])"
    NIV-->>EIK: "*"
    EIK-->>SGE: (stripped config, interrupt_kwargs)

    alt stream_mode includes events or JS graph
        SGE->>LG: "astream_events(input, config, **interrupt_kwargs)"
        LG-->>SGE: async event stream
    else standard streaming
        SGE->>LG: "astream(input, config, **interrupt_kwargs)"
        LG-->>SGE: async event stream
    end

    SGE-->>Caller: (mode, payload) tuples
```
</details>

<sub>Reviews (3): Last reviewed commit: ["Merge branch &#39;main&#39; into fix/run-level-i..."](https://github.com/aegra/aegra/commit/88fd94a962d1aff321f775a447579031a0ed555d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=42062582)</sub>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,10 @@ on:
   pull_request:
     branches: [main, develop]
 
+env:
+  # Keep in sync with the ruff floor in libs/*/pyproject.toml dev dependencies.
+  RUFF_VERSION: "0.16.0"
+
 jobs:
   lint:
     runs-on: ubuntu-latest
@@ -20,11 +24,13 @@ jobs:
       - name: Set up Python
         run: uv python install 3.12
 
+      # Pinned: an unpinned uvx pulls the newest ruff, so a release can turn
+      # every open PR red without a single line of repo code changing.
       - name: Check formatting with ruff
-        run: uvx ruff format --check .
+        run: uvx ruff@${{ env.RUFF_VERSION }} format --check .
 
       - name: Lint with ruff
-        run: uvx ruff check .
+        run: uvx ruff@${{ env.RUFF_VERSION }} check .
 
   test-api:
     runs-on: ubuntu-latest
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -3,7 +3,7 @@
   "info": {
     "title": "Aegra",
     "description": "Production-ready Agent Protocol server",
-    "version": "0.9.24"
+    "version": "0.9.25"
   },
   "paths": {
     "/info": {
--- a/libs/aegra-api/alembic/versions/20260612000000_hash_assistant_config_index.py
+++ b/libs/aegra-api/alembic/versions/20260612000000_hash_assistant_config_index.py
@@ -1,0 +1,73 @@
+"""Replace verbatim-config btree index with md5 expression index.
+
+The unique btree index ``idx_assistant_user_graph_config`` was introduced in
+revision ``76dfdbe90d2b`` (add_version_table) as::
+
+    CREATE UNIQUE INDEX idx_assistant_user_graph_config
+        ON assistant (user_id, graph_id, config);
+
+PostgreSQL btree index entries are capped at roughly **2 704 bytes** (one
+third of an 8 KB page). The ``config`` column is a JSONB object whose
+``configurable`` sub-key holds agent configuration — including a system
+prompt that routinely exceeds 2 704 bytes. Every ``INSERT`` or ``UPDATE``
+whose serialised ``config`` is larger than that limit fails with::
+
+    index row size NNNN exceeds btree version 4 maximum 2704 for index
+    "idx_assistant_user_graph_config"
+
+This migration replaces the verbatim-column btree with a fixed-width
+**expression index** over ``md5(config::text)``::
+
+    CREATE UNIQUE INDEX idx_assistant_user_graph_config
+        ON assistant (user_id, graph_id, md5(config::text));
+
+``md5(config::text)`` is always 32 bytes, so the index entry never exceeds
+the btree limit regardless of payload size.  The uniqueness guarantee is
+preserved: PostgreSQL stores JSONB in canonical form (normalised key order,
+no insignificant whitespace), so ``config::text`` is deterministic for
+logically-equal configs.  ``md5`` is a PostgreSQL built-in — no
+``pgcrypto`` extension is required.
+
+Downgrade note
+--------------
+The downgrade re-creates the original btree index.  This will fail on any
+database that already contains a row whose ``config::text`` exceeds
+~2 704 bytes (i.e. the exact problem this migration fixes).  Treat this
+migration as **forward-only** on production databases that have been used
+with large agent configs.
+
+Revision ID: b88bb61be638
+Revises: c7d1f2a4b6e8
+Create Date: 2026-06-12 00:00:00.000000
+"""
+
+from alembic import op
+
+revision = "b88bb61be638"
+down_revision = "c7d1f2a4b6e8"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.drop_index("idx_assistant_user_graph_config", table_name="assistant")
+    # CREATE INDEX cannot use CONCURRENTLY inside a transaction, and Alembic
+    # migrations run inside a transaction by default.  The brief ShareLock on
+    # the assistant table during the index build is acceptable: the table is
+    # small and this migration runs once at deployment time.
+    op.execute(
+        "CREATE UNIQUE INDEX idx_assistant_user_graph_config ON assistant (user_id, graph_id, md5(config::text))"
+    )
+
+
+def downgrade() -> None:
+    # WARNING: re-creating the verbatim-config btree will fail if any row's
+    # config::text already exceeds ~2 704 bytes.  Only safe on databases that
+    # have never stored a large-config assistant.
+    op.drop_index("idx_assistant_user_graph_config", table_name="assistant")
+    op.create_index(
+        "idx_assistant_user_graph_config",
+        "assistant",
+        ["user_id", "graph_id", "config"],
+        unique=True,
+    )
--- a/libs/aegra-api/pyproject.toml
+++ b/libs/aegra-api/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "aegra-api"
-version = "0.9.24"
+version = "0.9.25"
 description = "Aegra core API - Self-hosted Agent Protocol server"
 readme = "README.md"
 requires-python = ">=3.12"
--- a/libs/aegra-api/pyproject.toml
+++ b/libs/aegra-api/pyproject.toml
@@ -79,6 +79,11 @@ packages = ["src/aegra_api"]
 line-length = 120
 target-version = "py312"
 
+[tool.ruff.format]
+# This package's [tool.ruff] overrides the root config for files beneath it, so
+# the markdown exclusion has to be repeated here. See the root pyproject.toml.
+exclude = ["**/*.md"]
+
 [tool.ruff.lint]
 select = [
     "E",     # pycodestyle errors
@@ -119,7 +124,7 @@ dev = [
     "pytest-asyncio>=1.3.0",
     "pytest-mock>=3.15.1",
     "pytest-cov>=6.0.0",
-    "ruff>=0.13.0",
+    "ruff>=0.16.0",
     "ty>=0.0.17,<0.1.0",
     "bandit>=1.8.0",
     "pre-commit>=4.0.1",
--- a/libs/aegra-api/src/aegra_api/api/runs.py
+++ b/libs/aegra-api/src/aegra_api/api/runs.py
@@ -41,6 +41,24 @@ logger = structlog.getLogger(__name__)
 DEFAULT_STREAM_MODES = ["values"]
 
 
+async def _apply_create_run_auth(user: User, thread_id: str, request: RunCreate) -> None:
+    """Authorize threads.create_run and merge config/context overrides into request.
+
+    Handler-returned filter dict wins; otherwise fall back to in-place value mutations.
+    """
+    ctx = build_auth_context(user, "threads", "create_run")
+    value = {**request.model_dump(), "thread_id": thread_id}
+    filters = await handle_event(ctx, value)
+
+    source = filters if filters is not None else value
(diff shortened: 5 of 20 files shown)


Daniyal0100101:
Note on the version bump: CodeRabbit flagged the 0.9.25 changes as potentially out of scope, but I kept them intentionally because this repository's CLAUDE.md PR guidance says to always bump the version before creating a PR and to keep aegra-api/aegra-cli versions in sync. Happy to remove the version bump if maintainers prefer external bugfix PRs not to carry release metadata.

ibbybuilds (approved):
approving. the diagnosis is right: LangGraph reads interrupt_before/interrupt_after as explicit kwargs and ignores them as config keys, so run-level breakpoints set through the API never reached the graph.

the shallow copy before popping is the right call, it avoids mutating the caller config. and normalizing ["*"] back to the bare "*" sentinel is the kind of detail that would otherwise fail silently for all-nodes interrupts.

one gap, and i am merging anyway rather than holding this: it fixes the v1 producer only. stream_native_v3_events in native_stream.py passes config straight through with no extraction, so the same keys are still dropped on the v2 path. that matters because the v2 command handler comment explicitly promises interrupts are forwarded. i will open a follow-up issue for it rather than block a correct fix.

CI green across the board.

ibbybuilds:
merging this. the diagnosis is right, the tests are the correct shape (recording graph asserting exact kwargs), and it fixes a real user-facing break where HITL breakpoints were silently ignored.

one structural note for the record, not a blocker on this PR.

there is a round trip here that should not exist. run_executor.py builds the config and stuffs these keys in:

```python
config["interrupt_before"] = items if isinstance(items, list) else [items]
```

then this PR pops them back out and unwraps ["*"] to "*" again. so `_normalize_interrupt_value` exists only to undo a conversion the same codebase performed a few frames earlier. the values start typed on `job.behavior`, get flattened into an untyped dict, and get reconstructed downstream.

the cost is that the config dict is now a channel carrying things that are not config, and every producer has to remember to extract them. that contract is invisible, and it is already broken: the v2 native path does not extract, so interrupts are still dropped there (#481). a third streaming path added later will miss it too.

the durable fix is passing `job.behavior.interrupt_before` straight from the executor to both producers as kwargs, deleting the config stuffing and the extraction together. that closes #481 by construction instead of by remembering.

@Daniyal0100101 you have the most context here, so first refusal is yours if you want it. no pressure either way, happy to put it in the open pool. i have written it up as #482.

PR #465, opened 2026-07-20
feat: automate auth E2E tests in GitHub Actions

AayushSaini101:
## Summary

- Add a third E2E CI job (`e2e-auth`) that boots Aegra with JWT mock auth via `aegra.auth.json` and runs the auth suite
- Rename `@pytest.mark.manual_auth` → `@pytest.mark.auth_only` (aligned with `prod_only`)
- Add `make e2e-auth` + `docker-compose.auth.yml` for local runs
- Document the auth E2E mode in CLAUDE.md and contributing docs

Closes #242

## Test plan

- [x] `make e2e-auth` — 32 passed locally
- [ ] Confirm new **E2E (Auth Enabled)** job runs on this PR / `workflow_dispatch`
- [ ] Confirm existing `e2e-dev-mode` / `e2e-prod-mode` still ignore `manual_auth_tests`
- [ ] Spot-check marker: `pytest --markers | grep auth_only`

## Summary by CodeRabbit

* **New Features**
  * Added an authentication-enabled E2E mode running auth-only tests via a new `make e2e-auth` command, with dedicated server configuration and a docker-compose auth override.
  * Added an `e2e-auth` CI job that provisions services, waits for server readiness, and executes the auth-focused suite.
* **Tests**
  * Introduced the `auth_only` pytest marker and retagged auth E2E tests to run under it (skipped by default).
* **Documentation**
  * Updated contributing/testing docs and the auth E2E test README with new run commands and configuration details.
* **Chores**
  * Bumped API and CLI versions to 0.9.25.

<h3>Greptile Summary</h3>

This PR adds automated auth end-to-end coverage for the API. The main changes are:

- A new GitHub Actions auth E2E job.
- A local `make e2e-auth` target with Docker Compose auth overrides.
- A shared `aegra.auth.json` config for JWT mock auth.
- The auth test marker renamed to `auth_only`.
- Docs updated for running the auth suite.

<h3>Confidence Score: 5/5</h3>

This looks safe to merge.
- No blocking issues found in the changed code.
- The auth-only pytest commands now override the default marker exclusion.
- The local and CI auth targets both select the intended auth test marker.

<details><summary><h3>Important Files Changed</h3></summary>

| Filename | Overview |
|----------|----------|
| .github/workflows/e2e.yml | Adds the auth-enabled E2E job and selects `auth_only` tests with the default exclusion removed. |
| Makefile | Adds `make e2e-auth` with the same auth-only pytest selection used in CI. |
[...]

PR #467, opened 2026-07-25
feat(api): add select/values/extract to POST /threads/search

### LINKED ISSUE #330: [FEATURE] Thread search response shape (select, values, extract)
## Summary

`POST /threads/search` returns a thin response — `thread_id`, `status`, `metadata`, `user_id`, `created_at`, `updated_at`. Frontends that build conversation lists (sidebar with titles, previews, last-message snippets) cannot render anything useful from a single search call. They either:

- show raw UUIDs (poor UX, see #189), or
- issue 1+N requests (one search + one `GET /threads/{id}/state` per thread, see #303).

This issue tracks bringing the search response shape up to par so a single search call returns enough data for typical chat-UI rendering.

This **supersedes #189 and #303**. Filter/sort/pagination correctness is tracked separately in #302 / PR #322 and not in scope here.

## Goals

A client should be able to hit `POST /threads/search` with one call and:

1. Choose which fields come back (response projection)
2. Get latest state `values` per thread (e.g. messages, custom state keys)
3. Pull just specific JSON paths out of state/metadata/config without shipping the whole state blob

### 1. `select` field on the request

Whitelist of fields to return per thread. When omitted, default current shape (back-compat).

```json
{
  "select": ["thread_id", "status", "metadata", "values"]
}
```

Whitelist (proposed):
- `thread_id`, `status`, `created_at`, `updated_at`, `metadata`, `config`
- `values`, `interrupts` (require checkpoint join)

### 2. `values` populated when `select` includes it

Latest checkpoint state, joined into the response. Skipped (and fast-path preserved) when `select` omits `values`.

Implementation question to settle: per-thread `aget_state` calls (simple, slow) vs single batched fetch from `checkpoints` table (fast, requires bypassing checkpointer abstraction). Lean toward batched.

### 3. `extract` field for JSON-path projection

```json
{
  "select": ["thread_id"],
  "extract": {
    "title": "values.messages[0].content",
    "last_msg": "values.messages[-1].content"
  }
}
```

Returns:

```json
[
  {
    "thread_id": "...",
    "extracted": {
      "title": "what is rust?",
      "last_msg": "...because borrow checking..."
    }
  }
]
```

Path syntax: `key.subkey`, `key[0]`, `key[-1]`, `key.subkey[2].name`. No wildcards, no filters. Just navigation.

Validation:
- paths must start with one of: `values.`, `metadata.`, `config.`, `interrupts.`
- max 10 paths per request
- missing path → `null` in result, no 500
- malformed path → 422 at request time

## Acceptance

- existing callers without `select`/`extract` see no response shape change
- `select=["values"]` returns latest state per thread in one call
- `extract` resolves typed values without forcing full state in response
- frontends can render sidebar titles + previews from one search call (no 1+N)
- response size stays bounded (soft cap on per-thread `values` bytes, marker on truncation)
- thread ownership enforced at every layer including the checkpoint join

## Sequencing

Two PRs after PR #322 lands:

- **PR-B**: `select` + `values` join. Adds the response projection, the optional checkpoint join, the back-compat fast path. Closes #189.
- **PR-C** (on top of PR-B): `extract` JSON-path resolver + endpoint wiring. Closes #303 and reinforces #189 close for clients that want only specific paths.

## Out of scope

- filter/sort/pagination correctness — tracked in #302, PR #322
- TTL / checkpoint lifecycle — tracked in #288
- `values @> {...}` filter (state-side filter, not state-side projection) — defer until checkpoint indexing strategy is decided
- `ids` filter, `limit` cap changes — separate hardening pass

## Notes

#189 (response lacks `values`) and #303 (no `extract` option) are both closed by this work. Closing both with a pointer here so the conversation lives in one place.

AayushSaini101:
## Summary
- Add optional `select` projection and checkpoint `values`/`interrupts`/`config` join on `POST /threads/search` so chat UIs can render lists without 1+N `get_state` calls
- Add optional `extract` JSON-path map (aliases → paths under `values.`/`metadata.`/`config.` only; max 10; `interrupts.` rejected) returned as `extracted`
- Checkpoint values are serialized to JSON-shaped dicts before select/extract so LangChain message objects resolve correctly
- Keep the default (no `select`/`extract`) response shape unchanged; patch bump to 0.9.25

Closes #330

## Test plan
- [x] Unit: JSON-path validation/resolution + ThreadSearchService projection/join/truncation
- [x] Integration: `TestSearchThreads` select/extract validation and projection
- [x] E2E: `test_search_select_and_extract_values_e2e` against a running server
- [x] Confirm callers without `select`/`extract` still get the thin thread list

## Summary by CodeRabbit

* **New Features**
  * Enhanced `/threads/search` with response-field projection via `select` (returns only requested thread fields).
  * Added `extract` to compute derived fields from returned thread state, metadata, and configuration.
  * When `select`/`extract` are omitted, the default “thin” response shape is preserved.
* **Documentation**
  * Updated the “Searching threads” guide with a new “Response projection (`select` / `extract`)” section.
* **Bug Fixes**
  * Invalid `select` fields or malformed `extract` paths now fail validation (HTTP 422); missing/failed state loading falls back to safe empty defaults and large values are truncated.
* **Chores**
  * Version bumped to `0.9.25`.

<h3>Greptile Summary</h3>

Adds projected thread-search responses with optional checkpoint-state joins and JSON-path extraction.
- Adds `select` and `extract` request validation and response shaping.
- Adds bounded concurrent checkpoint loading, serialization, and value-size truncation.
- Documents the new search options and adds unit, integration, and end-to-end coverage.
- Bumps the API and CLI packages to version 0.9.25.

<h3>Confidence Score: 3/5</h3>

The PR is not yet safe to merge because large-state extraction remains incorrect and state-loading failures are still reported as valid empty state.

Checkpoint values are truncated before extraction resolves requested paths, while graph-loading, state-fetch, timeout, and conversion failures are collapsed into the same empty response used for genuinely absent state.

**Files Needing Attention:** libs/aegra-api/src/aegra_api/services/thread_search_service.py

| Filename | Overview |
|----------|----------|
| libs/aegra-api/src/aegra_api/api/threads.py | Routes projected searches through the new service while retaining the legacy response path when projection is omitted. |
| libs/aegra-api/src/aegra_api/models/threads.py | Adds validated `select` and `extract` request fields and documents enriched search results. |
| libs/aegra-api/src/aegra_api/services/thread_search_service.py | Implements concurrent checkpoint loading, serialization, projection, extraction, and truncation. |
| libs/aegra-api/src/aegra_api/utils/json_path.py | Adds validation and resolution for the supported restricted JSON-path syntax. |

</details>

```mermaid
sequenceDiagram
    participant Client
    participant API as POST /threads/search
    participant Search as ThreadSearchService
    participant Graph as LangGraph
    Client->>API: Search filters + select/extract
    API->>API: Query matching thread rows
    API->>Search: Build projected response
    alt Checkpoint fields required
        Search->>Graph: Load latest state
        Graph-->>Search: Snapshot
        Search->>Search: Serialize and project state
    end
    Search-->>API: Sparse result objects
    API-->>Client: Thread search results
```
</details>

<sub>Reviews (4): Last reviewed commit: ["fix(api): patch thread search monkeypatc..."](https://github.com/aegra/aegra/commit/672f80aa727944f77159d5c3f5dc4d74ed8c570e) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47221962)</sub>


ibbybuilds:
CI is unblocked and running now @AayushSaini101, first-time contributor approval was what was holding it. results below.

the lint failure is not yours. latest ruff started formatting markdown code blocks and flags CLAUDE.md on main, so every open PR is red on lint right now. i am fixing that separately, ignore it.

test-api has a real failure:

```
AttributeError: module aegra_api has no attribute services
AttributeError: module object at aegra_api.services has no attribute services
```

that is a monkeypatch target problem. looks like a patch path is being given as a string that resolves wrong, something like monkeypatch.setattr("aegra_api.services.services.x", ...) or patching an attribute on the package rather than the module that actually defines it. worth checking the patch targets in the new thread-search tests against how the existing service tests do it.

everything else passes, including codecov. also your e2e test test_search_select_and_extract_values_e2e is still unchecked in the description, worth running that locally before this goes in since the values join is the part most likely to surprise.

thanks for the patience on the CI approval, that was on me.

AayushSaini101:
> CI is unblocked and running now @AayushSaini101, first-time contributor approval was what was holding it. results below.
> 
> the lint failure is not yours. latest ruff started formatting markdown code blocks and flags CLAUDE.md on main, so every open PR is red on lint right now. i am fixing that separately, ignore it.
> 
> test-api has a real failure:
> 
> ```
> AttributeError: module aegra_api has no attribute services
> AttributeError: module object at aegra_api.services has no attribute services
> ```
> 
> that is a monkeypatch target problem. looks like a patch path is being given as a string that resolves wrong, something like monkeypatch.setattr("aegra_api.services.services.x", ...) or patching an attribute on the package rather than the module that actually defines it. worth checking the patch targets in the new thread-search tests against how the existing service tests do it.
> 
> everything else passes, including codecov. also your e2e test test_search_select_and_extract_values_e2e is still unchecked in the description, worth running that locally before this goes in since the values join is the part most likely to surprise.
> 
> thanks for the patience on the CI approval, that was on me.

thanks for reviewing,  i will take a look : )

AayushSaini101:
image
Tested unit tests and integration tests locally thanks

AayushSaini101:
@ibbybuilds can you review the changes thanks : )

ibbybuilds:
I approved the workflow runs on this so CI could actually execute, and reviewed the implementation. The design is good, but there are two things to fix before merge.

**Blocking: test-api fails.**

```
AttributeError: module aegra_api has no attribute services
AttributeError: module object at aegra_api.services has no attribute services
```

The doubled `services.services` in the second line points at a monkeypatch target that resolves one level deeper than intended. The string targets I can see in the diff (`aegra_api.services.thread_search_service.get_langgraph_service` and friends) look right, so it is likely one of the object-form patches (`monkeypatch.setattr(service, ...)`) or a target built by joining a module path. Worth running `pytest libs/aegra-api/tests/unit -q` locally, the failure reproduces without a database.

**Blocking: lint fails on `thread_search_service.py:66`.** Run `uv run ruff format .` and push. Ignore the three markdown files in the same output, those come from your branch predating #469 and clear themselves on a rebase.

**Design note, not blocking, but I want your answer on the record.** `_fetch_states_for_threads` issues one `aget_state` per thread behind an 8-way semaphore. That is N+1, which is the exact thing #330 asked us to avoid in favour of a batched fetch from the checkpoints table. I am not asking you to change it now, because the way you gated it is genuinely careful: no checkpoint work happens at all unless `select` or `extract` needs it, so the default search path is untouched. But it means a search returning 50 threads with `select=["values"]` does 50 round trips, and that will show up under load. Please either note in the PR description that batching is a deliberate follow-up, or open an issue for it, so it does not get silently forgotten.

**What is good:**

- The back-compat guard is right. When `select` and `extract` are both absent the response shape is byte-identical to today, so existing clients are unaffected.
- Fetching checkpoints only when the requested fields need them is the correct optimisation and shows you thought about the default path.
- The JSON path handling is properly locked down: prefix allowlist, validation at request time so malformed paths 422 instead of blowing up mid-response, and a 10-path cap. That matches the spec in #330 exactly.
- Splitting this into its own service plus a `json_path` util rather than growing the threads endpoint was the right call.

Fix the two CI failures and answer the N+1 question and this is ready. Thanks for the work, the shape of it is right.

ibbybuilds:
@AayushSaini101 heads up, CI had never actually run on this. Only the bots were reporting. I approved the workflow runs and two real failures showed up, details in my comment above.

Short version: `test-api` fails on a monkeypatch target resolving one level too deep (`services.services`), and `lint` needs `ruff format` on `thread_search_service.py`. Both reproduce locally with `pytest libs/aegra-api/tests/unit -q`, no database needed.

The feature itself is good and the security handling on the extract paths is right. Fix those two and answer the N+1 question and I will merge.

Separately, #465 is green and I am merging that one, thanks for it.

Issue #492, opened 2026-08-07
[BUG] aegra dev and aegra serve overwrite a pre-set AEGRA_CONFIG, so only -c can select an alternate manifest


alex-feel:
**Describe the bug**

`aegra dev` and `aegra serve` auto-discover `./aegra.json` without ever looking at the `AEGRA_CONFIG` environment variable, and then unconditionally overwrite that variable with the path they discovered. As a result, an `AEGRA_CONFIG` that the user exported in the shell — or set in the project's `.env` — is silently discarded and plain `aegra.json` is served. The explicit `-c/--config` flag is the only way to select a different manifest.

The overwrite lives in two places:

- `libs/aegra-cli/src/aegra_cli/cli.py` L303-L304 (`dev`): <https://github.com/aegra/aegra/blob/55744218fb18f7e1c50fb3ec799c5a59a9ac7bce/libs/aegra-cli/src/aegra_cli/cli.py#L303-L304>

  ```python
  # Set AEGRA_CONFIG env var so aegra-api resolves paths relative to config location
  os.environ["AEGRA_CONFIG"] = str(resolved_config)
  ```

- `libs/aegra-cli/src/aegra_cli/cli.py` L541-L542 (`serve`): <https://github.com/aegra/aegra/blob/55744218fb18f7e1c50fb3ec799c5a59a9ac7bce/libs/aegra-cli/src/aegra_cli/cli.py#L541-L542>

`resolved_config` comes from `find_config_file()` (L103-L119), which only probes `Path.cwd() / "aegra.json"` and then `Path.cwd() / "langgraph.json"` — it never reads `os.environ["AEGRA_CONFIG"]`.

The `.env` case fails for a second, compounding reason: `load_env_file()` only sets a key that is **not** already present in `os.environ` (`libs/aegra-cli/src/aegra_cli/env.py` L30-L33), and in `dev` the overwrite at L304 runs *before* `load_env_file()` at L322 (in `serve`, L542 before L546). So by the time the `.env` is parsed, `AEGRA_CONFIG` is already occupied by the auto-discovered path and the `.env` value is dropped.

This contradicts three things the project itself states:

1. `docs/reference/configuration.mdx` documents the resolution order as "1. `AEGRA_CONFIG` environment variable — absolute or relative path, 2. `aegra.json` in the current working directory, 3. `langgraph.json` …" and shows this exact example, which does not work:

   ```bash
   # Use a custom config file
[...]

Issue #490, opened 2026-08-05
[BUG] Thread search does not use build_metadata_filter, silently dropping the documented flat filter shape and every operator


connorimes:
**Describe the bug**

Aegra ships a filter compiler, `build_metadata_filter` in `aegra_api/core/auth_filters.py`, whose docstring defines the filter language an `@auth.on` handler may return: bare values for equality, `{"$eq": ...}`, `{"$contains": ...}`, `{"$or": [...]}`, `{"$and": [...]}`, multiple top-level keys as a conjunction, and — explicitly — a `{"metadata": {...}}` envelope that is flattened so that `{"metadata": {"team_id": "t"}}` and `{"team_id": "t"}` filter identically. It validates as it compiles, raising `HTTPException(500)` on an unknown operator rather than silently matching the wrong rows.

That compiler is called from exactly one file. Every one of its eight call sites is in `aegra_api/services/assistant_service.py`. No thread route, run route, cron route, or store route uses it.

`search_threads` in `aegra_api/api/threads.py` applies the handler's filter with its own code in the route. It checks `if filters and "metadata" in filters`, merges `filters["metadata"]` into the request's own metadata, and then matches with JSONB containment: `stmt.where(ThreadORM.metadata_json.op("@>")(request.metadata))`. Containment tests literal keys for equality, so the parts of the filter language that go beyond equality are not available on this route. Two shapes that the compiler handles correctly are mishandled here, both without any error.

A filter returned in the flat shape is ignored completely. `{"owner": "u1"}` — the shape `langgraph_sdk.Auth`'s own documentation shows, and the shape `build_metadata_filter` accepts — has no `"metadata"` key, so the `if` is false, nothing is merged, and the query proceeds with only `user_id == user.identity` for scoping. The handler's decision is discarded in full. This is the direction that matters most: a deployment that writes the documented handler, verifies it against assistants where it works, and assumes threads behave the same way has an unscoped thread search and no signal that anything is wrong.

A filter containing an operator matches nothing. `{"metadata": {"$or": [{"owner": "u1"}, {"visibility": "shared"}]}}` passes the `"metadata"` check and is merged, after which containment looks for a literal JSON key named `"$or"` in each thread's metadata column. No thread has one, so the result is an empty list rather than the union the handler asked for. `$and`, `$eq`, and `$contains` fail the same way. The compiler would have either compiled these correctly or rejected them with a 500; the containment check does neither.

[...]

Issue #485, opened 2026-08-03
[BUG] Auth handler filter is discarded on by-id thread/assistant read & delete routes


connorimes:
**Describe the bug**

Disclosure: Discovered and written with the help of Claude Sonnet 5 while planning to implement more complete auth in our agent.
Thanks for working on it!

An authorization handler registered via `@auth.on.threads.read` / `.delete` (or the equivalent `assistants.read` / `.delete`) can return a filter dict per the documented `Auth.types.FilterType` contract — the pattern shown in `langgraph_sdk.Auth`'s own docstring, e.g. `return {"owner": ctx.user.identity}` on `@auth.on.threads.read`. On Aegra's single-resource GET/DELETE routes, that return value is passed to `handle_event()` only to enforce a hard `False`/raised-exception outcome — the filter dict itself is never merged into the underlying query. A handler that scopes access by returning a filter therefore has no effect on `GET /threads/{thread_id}` or `DELETE /threads/{thread_id}`: any two callers sharing the same `user.identity` can fetch or delete each other's threads by id, regardless of what the handler returns.

The same gap exists on `GET /threads` (the plain list route, as distinct from `POST /threads/search`) and on the equivalent by-id `assistants` routes (`GET`/`DELETE /assistants/{assistant_id}`).

By contrast, `POST /threads/search`, `POST /assistants/search`, thread/assistant `create`, and thread/assistant `update` *do* apply the handler's returned filter (merged into `metadata` and used in the query), so the documented pattern works correctly there. Filtering is implemented for search/create/update and silently skipped for by-id read/delete/list.

**To Reproduce**

1. Configure a custom `auth.py` with:
   ```python
   from langgraph_sdk import Auth

   auth = Auth()

   @auth.authenticate
   async def authenticate(headers: dict[str, str]) -> dict[str, object]:
       # identity is per-caller-service here, not per-tenant — a common shape
       # when one trusted service credential acts on behalf of many tenants
       return {"identity": "service-a", "org_id": headers.get("x-org-id", "")}

   @auth.on.threads.read
   async def scope_thread_read(ctx: Auth.types.AuthContext, value):
       return {"metadata": {"org_id": ctx.user.org_id}}
   ```
2. Create a thread as org `A` (`POST /threads` with header `x-org-id: org-a`), and note its `thread_id`.
[...]

PR #494, opened 2026-08-07
fix(auth): dispatch @auth.on handlers on every protocol route

### LINKED ISSUE #488: [BUG] Authorization handlers are never invoked on thread state/history, several run routes, and stateless runs
**Describe the bug**

Aegra dispatches `@auth.on.*` authorization handlers by calling `handle_event()` explicitly inside each route. A number of routes never make that call, so no handler runs for them at all — not the specific handler for the resource and action, and not the catch-all `@auth.on` handler either. The consequence is stronger than a filter being ignored: on these routes a handler cannot deny a request even by returning `False` or raising an exception, because it is never given the request to consider. The deny-by-default pattern shown in `langgraph_sdk.Auth`'s own documentation — register `@auth.on` returning `False` first, then open specific resources with more specific handlers — does not hold on them. A deployment that follows that documented pattern and verifies it against `POST /threads`, `GET /threads/{thread_id}`, and `POST /threads/search` will reasonably conclude the runtime is closed, while the routes below remain open.

In `aegra_api/api/threads.py`, the routes with no authorization dispatch are `GET /threads/{thread_id}/state`, `POST /threads/{thread_id}/state`, `GET /threads/{thread_id}/state/{checkpoint_id}`, `POST /threads/{thread_id}/state/checkpoint`, `GET /threads/{thread_id}/history`, and `POST /threads/{thread_id}/history`. In `aegra_api/api/runs.py` they are `GET /threads/{thread_id}/runs`, `PATCH /threads/{thread_id}/runs/{run_id}`, `GET /threads/{thread_id}/runs/{run_id}/join`, `GET /threads/{thread_id}/runs/{run_id}/stream`, and `POST /threads/{thread_id}/runs/{run_id}/cancel`. In `aegra_api/api/stateless_runs.py` no route calls `build_auth_context` or `handle_event` at all, so `POST /runs`, `POST /runs/wait`, and `POST /runs/stream` execute a graph and create and delete an ephemeral thread without any authorization handler being consulted.

This is worth weighing against the routes that do dispatch. `create_thread`, `list_threads`, `get_thread`, `update_thread`, `delete_thread`, `search_threads`, `get_run`, `delete_run`, the store routes, and the cron routes all call `handle_event()`, as does run creation via `_apply_create_run_auth`. So authorization is dispatched on most of the surface and skipped on a subset — and the subset happens to include the routes that return conversation content. `GET /threads/{thread_id}` is authorized; `GET /threads/{thread_id}/state` and `GET /threads/{thread_id}/history`, which return the messages that thread holds, are not.

This is distinct from and compounds https://github.com/aegra/aegra/issues/485. That issue concerns routes that *do* dispatch a handler and then discard the filter it returns, so a handler can still deny outright there, and fixing it makes returned filters effective on by-id read, delete, and the plain list. Neither the issue nor its fix reaches the routes described here, because there is no dispatch to attach a filter to. A deployment relying on `@auth.on` for isolation between principals that share an `identity` value would still have unfiltered access to thread state and history after #485 is resolved.

**To Reproduce**

1. Configure a custom `auth.py` that follows the documented deny-by-default pattern, opening only what is needed to get a thread and a run into existence:
   ```python
   from typing import Any

   from langgraph_sdk import Auth

   auth = Auth()

   @auth.authenticate
   async def authenticate(headers: dict[str, str]) -> dict[str, object]:
       return {"identity": "user-a", "is_authenticated": True}

   @auth.on
   async def deny_by_default(ctx: Auth.types.AuthContext, value: Any) -> bool:
       return False

   @auth.on.threads.create
   async def allow_thread_create(ctx: Auth.types.AuthContext, value) -> None:
       pass

   @auth.on.threads.create_run
   async def allow_run_create(ctx: Auth.types.AuthContext, value) -> None:
       pass
   ```
2. `POST /threads` and `POST /threads/{thread_id}/runs` — both succeed, as intended.
3. `GET /threads/{thread_id}` — denied, confirming the catch-all is registered and effective.
4. `GET /threads/{thread_id}/state` and `GET /threads/{thread_id}/history` — both return 200 with the thread's state and full checkpoint history, despite the same catch-all that denied step 3. `GET /threads/{thread_id}/runs` and `POST /runs/stream` behave the same way.

This can also be confirmed by reading the source directly: `handle_event` appears only at three call sites in `aegra_api/api/runs.py` (`_apply_create_run_auth`, `get_run`, `delete_run`), at six in `aegra_api/api/threads.py` (`create_thread`, `list_threads`, `get_thread`, `update_thread`, `delete_thread`, `search_threads`), and nowhere in `aegra_api/api/stateless_runs.py`. The routes listed above query the database directly with `user.identity` as their only predicate.

**Expected behavior**

Every route that reads or mutates a resource dispatches the corresponding authorization event before touching the database, so that a registered handler — specific or catch-all — is given the opportunity to deny the request or to scope it. Concretely, the thread state and history GET routes would dispatch `threads.read`, `POST /threads/{thread_id}/state` would dispatch `threads.update`, the run list/join/stream routes would dispatch `runs.read`, the run PATCH and cancel routes would dispatch `runs.update`, and the stateless run routes would dispatch `threads.create` and `threads.create_run` as their stateful equivalents already do. A deployment that registers `@auth.on` returning `False` and nothing else would then have every route closed.

**Actual behavior**

The routes listed above are served with `user.identity` as the sole access-control predicate, and no authorization handler participates in the decision. A handler that returns `False` for those resource/action pairs has no effect, and a handler that returns a filter has nothing to apply it to.

**Environment:**
- OS: macOS 14.x (Darwin 24.6.0)
- Python version: 3.12
- Aegra version: 0.9.25
- LangGraph version: langgraph 1.2.10, langgraph-sdk 0.4.2

**Configuration:**
- Database: PostgreSQL
- Auth type: custom
- Deployment: local

**Logs/Error Messages:**
```
N/A — not an exception. The bug is silently permissive rather than erroring.
```

**Additional context**

The 0.9.25 changelog entry "call auth handlers in stream and wait run endpoints" addresses part of this class of gap, and the fix is visible in the source: `wait_for_run` and `create_and_stream_run` both call `_apply_create_run_auth`. That covers the routes that *create* a run. The read-side attach routes — `GET /threads/{thread_id}/runs/{run_id}/stream` and `.../join`, which stream the output of a run created earlier — were not part of that change and still have no dispatch, as do the thread state and history routes and the stateless run routes.

`handle_event`'s own docstring describes the intended posture as non-interruptive by default: no auth configured or no handlers defined means allow, and only a `False` return or a raised exception interrupts. That is a reasonable default and is not what this report disputes. The issue is that on these routes the handler is not consulted at all, so a deployment that *has* defined handlers gets the no-handlers-defined behavior on part of its surface without any signal that it has.

**Possible Solution**

Add the missing `build_auth_context` / `handle_event` calls to the routes listed above, mapping each to the resource and action its stateful equivalent already uses. Since the handler's returned filter would then be available on the thread state and history routes, it would be worth applying it there in whatever query-merging shape #485 settles on, so the two fixes leave a consistent contract rather than one where a filter is honored on `GET /threads/{thread_id}` but dropped on `GET /threads/{thread_id}/history`.

Thanks!
Disclosure: identified and written with help from Claude Opus 5.

ibbybuilds:
Closes #488, #490, #485, #492.

## The bug

Authorization was wired by hand inside each route body. A route that forgot the `handle_event` call was silently unprotected: no error, no failing test, the handler just never ran.

Verified against a running server with a handler that denies `threads.read`:

| Route | Before | After |
|---|---|---|
| `GET /threads/{id}` | 403 | 403 |
| `GET /threads/{id}/state` | **200, served the conversation** | 403 |
| `GET /threads/{id}/history` | **200, served the conversation** | 403 |

18 routes reached the database without dispatching: thread state/history/checkpoint, `list_runs`, `update_run`, `join_run`, `stream_run`, `cancel_run`, cron creation, and both v2 event-streaming routes.

To be precise about scope, since #488 reads wider than what I found: assistants are **not** affected. They dispatch through the `Authenticated` service base. Stateless runs are also fine, delegating to the threaded run functions. Tenant isolation via `user_id` was never broken; what broke is that a user's own `@auth.on` rules were ignored on those 18 routes.

## The fix

Route to `(resource, action)` now lives in `core/auth_registry.py` and is attached at registration, so a route opts out only by being listed exempt. `test_auth_registry.py` fails when a mounted route is neither registered nor exempt, which is what stops this from recurring.

- Sub-resources authorize as their parent, so `@auth.on.threads.read` covers state and history.
- Stateless runs authorize as `threads/create_run`, so dropping the `thread_id` cannot dodge an `@auth.on.threads` rule.
- Routes that already dispatch in-body are unchanged. The request is marked once authorized, so a handler still runs exactly once (covered by a test).

### Thread filters (#490, #485)

Thread search accepted only a nested `metadata` key and dropped the flat shape and every operator. `list_threads` was worse: it computed the filter and threw it away with `if filters: pass`. Both now compile through `build_metadata_filter`, as `assistant_service` already did, and by-id read and delete honor it too.

### CLI (#492)

`aegra dev` and `aegra serve` overwrote a pre-set `AEGRA_CONFIG` unconditionally. Discovery now honors it, warns and falls back if the path is stale, and treats a blank value as unset.

## Verification

- 1843 unit + integration pass
- 134 E2E pass in prod mode (Redis workers), 117 in dev mode
- Disabling the enforcement wiring makes 12 of the new tests fail, so they measure the fix rather than the mock
- `ruff check` and `format --check` clean

One pre-existing failure, `test_assistant_large_config_db.py`, fails identically on `main` in my environment. It needs `POSTGRES_*` set, unrelated to this change.

## Note for review

The enforcer depends on `require_auth`, not `get_current_user`. It is prepended, so it runs before the route's own dependencies, and `get_current_user` only reads the scope that `require_auth` populates. Getting this wrong 500s every request, and only the real-server run caught it, so there is now a test pinning the dependency.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Added consistent authorization coverage across Agent Protocol routes, including nested resources, streaming, stateless runs, and request metadata.
  - Thread searches and operations now honor complete authorization filters while preserving user-scope restrictions.
  - Added support for selecting a local CLI configuration file through the `AEGRA_CONFIG` environment variable.

- **Documentation**
  - Documented route authorization behavior, filtering, metadata handling, and validation.

- **Bug Fixes**
  - Improved configuration discovery fallback when the configured file is missing, blank, outside the working directory, or invalid.

<h3>Greptile Summary</h3>

The PR centralizes Agent Protocol authorization dispatch, expands thread-filter enforcement on direct thread operations, aligns run authorization with thread handlers, and makes CLI configuration discovery honor `AEGRA_CONFIG`.
- Adds a route-to-authorization registry and registration-time FastAPI dependencies.
- Applies metadata authorization filters to thread list, search, read, and delete queries.
- Adds assistant authorization during run creation and broad route-coverage regression tests.
- Updates CLI configuration discovery, documentation, and package versions.

<h3>Confidence Score: 2/5</h3>

The PR is not yet safe to merge because dependency-enforced routes still ignore authorization filters, and body-route handler mutations remain ineffective.

The central enforcer calls authorization handlers but discards both returned filter dictionaries and mutations to its temporary value object; state, history, run, and event routes then proceed using only identity-level ownership checks or separately parsed request models.

**Files Needing Attention:** libs/aegra-api/src/aegra_api/core/auth_enforcement.py, libs/aegra-api/src/aegra_api/api/threads.py, libs/aegra-api/src/aegra_api/api/runs.py, libs/aegra-api/src/aegra_api/api/event_streaming.py

| Filename | Overview |
|----------|----------|
| libs/aegra-api/src/aegra_api/core/auth_enforcement.py | Adds centralized authorization dependencies and fixes path-parameter precedence and exception handling, but still discards returned filters and in-place mutations. |
| libs/aegra-api/src/aegra_api/core/auth_registry.py | Defines comprehensive route mappings and exemptions, while relying on the enforcer and self-dispatch classifications to preserve each handler’s full contract. |
| libs/aegra-api/src/aegra_api/api/threads.py | Correctly applies authorization filters to direct thread list/search/read/delete queries, while state and history routes still lack equivalent filter enforcement. |
| libs/aegra-api/src/aegra_api/api/runs.py | Aligns run events with thread authorization and adds assistant checks during creation, but dependency-enforced run operations do not apply returned thread filters. |
| libs/aegra-api/src/aegra_api/api/event_streaming.py | Registry enforcement now dispatches handlers for v2 routes, but their returned filters and request mutations are not propagated. |
| libs/aegra-api/src/aegra_api/main.py | Applies registry-based authorization after mounting core routers for both stock and custom FastAPI applications. |
| libs/aegra-cli/src/aegra_cli/cli.py | Updates configuration discovery to respect valid `AEGRA_CONFIG` values and provide fallback behavior for stale or blank values. |

</details>

```mermaid
sequenceDiagram
    participant Client
    participant Enforcer as Auth enforcer dependency
    participant Handler as @auth.on handler
    participant Route
    participant DB
    Client->>Enforcer: Protocol request
    Enforcer->>Handler: handle_event(ctx, temporary value)
    Handler-->>Enforcer: Allow or metadata filter
    Note over Enforcer: Returned filter is discarded
    Enforcer->>Route: Continue request
    Route->>DB: Query by resource IDs and user_id
    DB-->>Route: Resource outside handler metadata scope
    Route-->>Client: Data or mutation result
```
</details>

<sub>Reviews (8): Last reviewed commit: ["chore: bump version to 0.10.1"](https://github.com/aegra/aegra/commit/6ca9ab51febf855c78b2a45b9525c1d9594ccd07) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51228408)</sub>

<details><summary><h4>Context used (3)</h4></summary>

- Knowledge Base — [Authentication and Authorization](https://app.greptile.com/aegra/-/custom-context/knowledge-base/aegra/aegra/-/docs/auth.md)
- Knowledge Base — [App core & server bootstrap](https://app.greptile.com/aegra/-/custom-context/knowledge-base/aegra/aegra/-/docs/app-core-server.md)
</details>

### DIFF
--- a/docs/guides/authentication.mdx
+++ b/docs/guides/authentication.mdx
@@ -178,6 +178,15 @@ Stateless runs create an ephemeral thread, so they authorize as
 A filter dict returned from a `threads` handler is applied to thread queries,
 including by-id reads and deletes, not only to search.
 
+Your handler runs **once per event**. A request that legitimately touches
+several resources still checks each one: creating a cron authorizes
+`crons.create`, then `assistants.read`, then the thread, so you can deny at any
+layer.
+
+Handlers may also inject data by mutating `value` in place. Setting
+`value["metadata"]["team_id"]` in an `@auth.on.threads.create` handler stores
+that metadata on the new thread.
+
 <Note>
 Route-to-handler mapping lives in `core/auth_registry.py`. A route that is not
 registered there fails the test suite, so a new endpoint cannot silently skip
--- a/docs/guides/authentication.mdx
+++ b/docs/guides/authentication.mdx
@@ -165,12 +165,14 @@ as their parent, so one rule covers a whole family of endpoints:
 
 | Handler | Also covers |
 |---------|-------------|
-| `@auth.on.threads.read` | `GET /threads/{id}`, `/state`, `/state/{checkpoint_id}`, `/history` |
-| `@auth.on.threads.update` | `PATCH /threads/{id}`, `POST /threads/{id}/state` |
+| `@auth.on.threads.read` | `GET /threads/{id}`, `/state`, `/state/{checkpoint_id}`, `/history`, `GET` a run, its `/join` and `/stream` |
+| `@auth.on.threads.update` | `PATCH /threads/{id}`, `POST /threads/{id}/state`, `PATCH` a run, `POST .../runs/{id}/cancel` |
+| `@auth.on.threads.delete` | `DELETE /threads/{id}`, `DELETE` a run |
 | `@auth.on.threads.create_run` | `POST /threads/{id}/runs`, `/runs/stream`, `/runs/wait`, `/threads/{id}/commands`, and the stateless `POST /runs`, `/runs/stream`, `/runs/wait` |
-| `@auth.on.threads.search` | `GET /threads`, `POST /threads/search` |
-| `@auth.on.runs.read` | `GET` a run, its `/join` and `/stream` |
-| `@auth.on.runs.update` | `PATCH` a run, `POST /runs/{id}/cancel` |
+| `@auth.on.threads.search` | `GET /threads`, `POST /threads/search`, `GET /threads/{id}/runs` |
+
+There is no `runs` resource: the Agent Protocol authorizes run operations under
+the thread that owns them, and the SDK's `@auth.on` has no `runs` attribute.
 
 Stateless runs create an ephemeral thread, so they authorize as
 `threads.create_run`. Omitting the `thread_id` does not bypass your thread rules.
@@ -181,7 +183,8 @@ including by-id reads and deletes, not only to search.
 Your handler runs **once per event**. A request that legitimately touches
 several resources still checks each one: creating a cron authorizes
 `crons.create`, then `assistants.read`, then the thread, so you can deny at any
-layer.
+layer. Creating a run likewise authorizes `threads.create_run`, then
+`assistants.read` for the assistant it uses.
 
 Handlers may also inject data by mutating `value` in place. Setting
 `value["metadata"]["team_id"]` in an `@auth.on.threads.create` handler stores
--- a/docs/guides/authentication.mdx
+++ b/docs/guides/authentication.mdx
@@ -158,6 +158,32 @@ The `ctx` parameter provides:
 | `dict` | Allow with filters (e.g., `{"metadata": {"team_id": "123"}}`) |
 | Modified `value` | Allow with modified request data |
 
+### Which routes a handler covers
+
+Every Agent Protocol route dispatches to your handlers. Sub-resources authorize
+as their parent, so one rule covers a whole family of endpoints:
+
+| Handler | Also covers |
+|---------|-------------|
+| `@auth.on.threads.read` | `GET /threads/{id}`, `/state`, `/state/{checkpoint_id}`, `/history` |
+| `@auth.on.threads.update` | `PATCH /threads/{id}`, `POST /threads/{id}/state` |
+| `@auth.on.threads.create_run` | `POST /threads/{id}/runs`, `/runs/stream`, `/runs/wait`, `/threads/{id}/commands`, and the stateless `POST /runs`, `/runs/stream`, `/runs/wait` |
+| `@auth.on.threads.search` | `GET /threads`, `POST /threads/search` |
+| `@auth.on.runs.read` | `GET` a run, its `/join` and `/stream` |
+| `@auth.on.runs.update` | `PATCH` a run, `POST /runs/{id}/cancel` |
+
+Stateless runs create an ephemeral thread, so they authorize as
+`threads.create_run`. Omitting the `thread_id` does not bypass your thread rules.
+
+A filter dict returned from a `threads` handler is applied to thread queries,
+including by-id reads and deletes, not only to search.
+
+<Note>
+Route-to-handler mapping lives in `core/auth_registry.py`. A route that is not
+registered there fails the test suite, so a new endpoint cannot silently skip
+your handlers.
+</Note>
+
 ## Accessing user data in your graph
 
 When auth is enabled, Aegra automatically injects the authenticated user's data into
--- a/libs/aegra-api/pyproject.toml
+++ b/libs/aegra-api/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "aegra-api"
-version = "0.10.0"
+version = "0.10.1"
 description = "Aegra core API - Self-hosted Agent Protocol server"
 readme = "README.md"
 requires-python = ">=3.12"
--- a/libs/aegra-api/src/aegra_api/api/runs.py
+++ b/libs/aegra-api/src/aegra_api/api/runs.py
@@ -196,8 +196,9 @@ async def get_run(
     Returns the current state of the run including its status, input, output,
     and error information.
     """
-    # Authorization check (read action on runs resource)
-    ctx = build_auth_context(user, "runs", "read")
+    # Reading a run authorizes as a thread read (Agent Protocol has no `runs`
+    # resource); @auth.on.threads.read covers it.
+    ctx = build_auth_context(user, "threads", "read")
     value = {"run_id": run_id, "thread_id": thread_id}
     await handle_event(ctx, value)
 
@@ -539,8 +540,9 @@ async def delete_run(
     Conflict. Set `force=1` to cancel the run first (best-effort) and then
     delete it. Returns 204 No Content on success.
     """
-    # Authorization check (delete action on runs resource)
-    ctx = build_auth_context(user, "runs", "delete")
+    # Deleting a run authorizes as a thread delete (Agent Protocol has no `runs`
+    # resource); @auth.on.threads.delete covers it.
+    ctx = build_auth_context(user, "threads", "delete")
     value = {"run_id": run_id, "thread_id": thread_id}
     await handle_event(ctx, value)
     logger.info(f"[delete_run] fetch run run_id={run_id} thread_id={thread_id} user={user.identity}")
--- a/libs/aegra-api/src/aegra_api/api/runs.py
+++ b/libs/aegra-api/src/aegra_api/api/runs.py
@@ -100,6 +100,13 @@ async def _apply_create_run_auth(user: User, thread_id: str, request: RunCreate)
     if isinstance(context_overrides, dict):
         request.context = {**(request.context or {}), **context_overrides}
 
+    # Creating a run also reads its assistant, so per-assistant handler rules
+    # apply here exactly as they do in the cron-create chain.
+    await handle_event(
+        build_auth_context(user, "assistants", "read"),
+        {"assistant_id": request.assistant_id},
+    )
+
 
 @router.post("/threads/{thread_id}/runs", response_model=Run, responses={**NOT_FOUND, **CONFLICT})
 async def create_run(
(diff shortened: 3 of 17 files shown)


ibbybuilds:
Second pass after review. The first version fixed the bypass but introduced two quieter ones, both from the same root cause: two layers sharing one dispatch, with a dedup flag between them.

**1. Multi-resource requests collapsed to their first check.** Creating a cron authorizes `crons.create`, then `assistants.read`, then the thread. That chain is a documented contract (`crons.py`), so a caller with cron access but no rights to the underlying assistant gets refused. A single "already dispatched" flag stopped after the first event, so the assistant and thread checks silently stopped running.

**2. Handler metadata injection was dropped.** Handlers inject by mutating `value` in place, which the shipped `jwt_mock_auth_example.py` does for `team_id`, `org_id` and `created_by`. The enforcer ran the handler against a throwaway dict and skipped the route's own call, so nothing reached the request model.

`make e2e-auth` caught the second one. That suite exercises real handlers end to end and I had not run it. Thread create came back with `team_id` unset.

**Fix:** exactly one layer authorizes each event. Routes in `SELF_DISPATCHING` keep their in-body call and get no enforcer, since their `value` is the live request model. Everything else is dispatched centrally. The dedup guard in `handle_event` is deleted rather than made smarter, because nothing dispatches twice any more.

Also: read the body on DELETE (`/store/items` carries one), and narrow the body catch to `JSONDecodeError`/`UnicodeDecodeError`.

**New tests**, each failing before this commit: the cron chain reaches every layer, thread create keeps injected metadata, and the store delete handler sees its namespace and key. `SELF_DISPATCHING` is also checked against the registry so a stale entry cannot quietly disable the enforcer.

**Verification**
- 1839 unit + integration pass
- 134 e2e in prod mode (Redis workers)
- auth e2e back to its pre-branch baseline: 30 passed, 2 failed. Those 2 reproduce on `main` with no local changes and are filed as #495
- re-confirmed the original bypass stays closed against a live server: `/threads/{id}`, `/state` and `/history` all 403 under a handler denying `threads.read`

The original defect and both regressions were real. Worth noting the auth e2e job is not currently a clean signal because of #495.

ibbybuilds:
Third commit. CI caught a real bug in the #492 fix that I had missed locally.

`test-cli` was hanging and getting killed after 10 minutes. Root cause: the first version honored any `AEGRA_CONFIG` that existed on disk, so an env var exported in a shell leaked into every later `aegra dev`. `test_dev_fails_without_config` runs in an empty directory and expects exit 1, but discovery now succeeded, so the CLI started a server and pytest never returned.

That is a genuine usability flaw, not just a test artifact: a stale `AEGRA_CONFIG` would silently boot a different project than the directory you are standing in.

Now the env var is honored only when it resolves inside the current directory. The reported case still works, since that container sets `AEGRA_CONFIG=aegra.json` relative to its workdir. An unrelated absolute path falls back to discovery, and `-c` remains the way to point outside the tree. Verified both directions, including the empty-directory case that hung.

All checks green: `lint`, `test-api`, `test-cli`, `CodeQL`, `codecov/patch`, plus Greptile and CodeRabbit.

Worth noting how this surfaced. The local CLI suite hangs the same way on any machine with `AEGRA_CONFIG` set, so I could not see it until CI ran the job to completion. Same shape as the earlier one: the failure lived exactly where I had not run the tests.

connorimes:
Thanks for working on these issues. I started working on a couple of them, too, before I saw this PR. I didn't finish with the dev-mode E2E test failures I was getting (probably due to lack of an OpenAI API key). I figured I'd stop there since you're already working on these, but it was a useful exercise to see what's involved in contributing to Aegra. For reference, in case you find anything helpful from the work (using Claude Opus 5): https://github.com/connorimes/aegra/tree/fix/485-thread-auth-filters

Cheers.

Issue #288, opened 2026-04-04
[FEATURE] Thread & Checkpoint Lifecycle Management (TTL + Cascade Cleanup)


ibbybuilds:
**Is your feature request related to a problem? Please describe.**

When a thread is deleted in Aegra (via `DELETE /threads/{id}` or ephemeral thread cleanup in stateless runs), the thread row and its runs are removed via FK cascade. However, the checkpoint data (`checkpoints`, `checkpoint_blobs`, `checkpoint_writes` tables) managed by the checkpointer is **not cleaned up**. This leaves orphaned checkpoint data in the database indefinitely.

Over time — especially in production deployments with high thread volume — this leads to:
- Unbounded database growth from dead checkpoint data
- Increased storage costs
- Potential query performance degradation

Additionally, there is no automatic lifecycle management for threads. Threads created by users or stateless runs live forever unless manually deleted, requiring external cron jobs or manual database pruning.

**Describe the solution you'd like**

A two-phase approach:

### Phase 1 — Fix orphaned checkpoints on thread deletion (bug fix)

Every code path that deletes a thread should also call `await checkpointer.adelete_thread(thread_id)` to clean up associated checkpoint data. The checkpointer already exposes this method — it deletes from all three checkpoint tables for the given thread_id.

**Affected code paths:**
1. `DELETE /threads/{thread_id}` — `api/threads.py` (`delete_thread`)
2. `_delete_thread_by_id()` — `api/stateless_runs.py` (ephemeral thread cleanup)

**Implementation note:** The API layer currently doesn't have access to the checkpointer instance (it lives in `DatabaseManager`). This needs to be exposed — either via dependency injection, a service function, or a module-level accessor.

### Phase 2 — Configurable TTL with background sweep (new feature)

Automatic thread lifecycle management with configurable time-to-live.

**A. New `thread_ttl` database table** (Alembic migration):
```sql
CREATE TABLE thread_ttl (
    thread_id TEXT PRIMARY KEY REFERENCES thread(thread_id) ON DELETE CASCADE,
    strategy TEXT NOT NULL DEFAULT 'delete',
    ttl_minutes FLOAT NOT NULL,
    created_at TIMESTAMP NOT NULL DEFAULT (NOW() AT TIME ZONE 'UTC'),
    expires_at TIMESTAMP NOT NULL
);
CREATE INDEX idx_thread_ttl_expires_at ON thread_ttl(expires_at);
```

[...]

PR #499, opened 2026-08-09
fix(api): delete checkpoint data when a thread is deleted

arturbagramyan1:
## Summary

Phase 1 of #288. Deleting a thread never called the checkpointer's `adelete_thread`, so every deletion left orphaned rows in `checkpoints`, `checkpoint_blobs`, and `checkpoint_writes` forever (these tables have no FK to `thread`, so no DB cascade can clean them). Both deletion paths now clean up checkpoint data:

- `DELETE /threads/{thread_id}` — `api/threads.py::delete_thread`
- `services/run_cleanup.py::delete_thread_by_id` — the helper the issue refers to as `_delete_thread_by_id` in `stateless_runs.py` moved here since, and is shared by stateless runs, crons, and the cron scheduler — so this covers every ephemeral-thread cleanup path (the biggest orphan source, since ephemeral threads leak per run).

## Ordering decision

The checkpointer runs on the psycopg pool (autocommit), a separate transaction from the SQLAlchemy session, so the two deletes can't be atomic — ordering is the only lever. I picked **checkpoints first, thread-row commit second**:

| | `adelete_thread` fails | commit fails after checkpoints deleted |
|---|---|---|
| checkpoints → commit (chosen) | thread row intact, 500 returned, retry works, no orphans | thread survives as a valid empty thread; retrying the delete completes it |
| commit → checkpoints | row gone, checkpoints orphaned forever, retry 404s — reproduces the exact bug | — |

The reverse order has an unrecoverable failure mode; the chosen order's worst case is a benign, retryable partial state.

`adelete_thread` takes only a `thread_id`, so it is called strictly after the user-scoped ownership lookup in both paths — calling it without that check could wipe another user's checkpoints.

## Error handling

- API route: failures propagate → 500, row uncommitted, client retries.
- `run_cleanup`: `psycopg.Error` added to `_CLEANUP_ERRORS` so transient checkpoint-backend failures follow the existing tolerated-infra contract of the cleanup callers. Programmer errors (including `RuntimeError` from an uninitialized `db_manager`) still propagate.

## Tests

- **Unit** (`test_stateless_runs.py`): delete awaited with the right id; ordering regression (checkpoints strictly before commit); failure keeps the thread row; ownership guard skips the call when the row isn't found; widened-tuple swallow test for background cleanup.
- **Integration** (`test_threads_crud.py`): happy path calls `adelete_thread`; checkpoint failure → 500 with no row delete/commit; 404 path never touches the backend.
- **E2E** (`test_thread_deletion_removes_checkpoints`): runs the hermetic `stress_test` graph, asserts checkpoint rows exist, deletes the thread, asserts all three tables are empty for that thread. Verified in both dev (`LocalExecutor`) and prod (Redis workers) modes.

Also verified against a live stack: after exercising both deletion paths, `SELECT count(*) FROM checkpoints c WHERE NOT EXISTS (SELECT 1 FROM thread t WHERE t.thread_id = c.thread_id)` → 0.

## Known limitations

- In prod mode a worker mid-write can persist a checkpoint after `adelete_thread` ran (run cancellation via Redis is asynchronous) → a small bounded orphan. Inherent without cross-store transactions; Phase 2's TTL sweep is the designed backstop.
- A background cleanup task racing shutdown can hit `RuntimeError` from `get_checkpointer()` after `db_manager.close()`. Deliberately not added to the tolerated tuple so lifecycle bugs surface.

## Docs

- `delete_thread` docstring updated (it previously documented that checkpoint history is *not* affected) + `docs/openapi.json` regenerated, `threads-and-state.mdx` updated.

Phase 2 (TTL + background sweep per #288) follows as a separate PR once this lands.

cc @ibbybuilds

## Summary by CodeRabbit

* **Bug Fixes**
  * Deleting a thread now also removes its checkpoint history.
  * Thread records are preserved when checkpoint cleanup fails, allowing deletion to be retried safely.
  * Cleanup failures return an error instead of causing partial deletion.
  * Background cleanup tolerates transient database errors.
  * Active runs continue to be cancelled automatically before thread deletion.
* **Documentation**
  * Updated thread deletion guidance and API documentation to clarify checkpoint removal and failure handling.

<h3>Greptile Summary</h3>

This PR extends explicit and ephemeral thread deletion to remove LangGraph checkpoint history before committing the thread-row deletion, preserving retryability when checkpoint cleanup fails.
- Adds checkpoint cleanup to the API and shared run-cleanup paths.
- Documents the revised deletion behavior and API error response.
- Adds unit, integration, and end-to-end regression coverage.

<h3>Confidence Score: 4/5</h3>

The PR is not yet safe to merge because a transient checkpoint-backend failure permanently abandons the only scheduled cleanup for ephemeral threads.

The background cleanup path catches checkpoint-related database errors, logs them, and completes without retrying or rescheduling; since no orphan-thread sweeper currently exists, the thread and checkpoint rows remain indefinitely.

**Files Needing Attention:** libs/aegra-api/src/aegra_api/services/run_cleanup.py

| Filename | Overview |
|----------|----------|
| libs/aegra-api/src/aegra_api/api/threads.py | Adds ownership-gated checkpoint deletion before committing explicit thread deletion. |
| libs/aegra-api/src/aegra_api/services/run_cleanup.py | Adds checkpoint deletion to ephemeral cleanup, but a transient checkpoint error remains terminal because the sole background attempt only logs it. |
| libs/aegra-api/tests/e2e/test_threads/test_thread_deletion.py | Adds end-to-end verification that all three checkpoint tables are cleared using literal parameterized queries. |
| libs/aegra-api/tests/integration/test_api/test_threads_crud.py | Covers successful checkpoint deletion, ownership-safe not-found behavior, and explicit-route failure ordering. |
| libs/aegra-api/tests/unit/test_api/test_stateless_runs.py | Covers ephemeral cleanup ordering and error handling, including the behavior that leaves transient failures without another cleanup attempt. |

</details>

```mermaid
sequenceDiagram
    participant Caller
    participant Cleanup as Thread cleanup
    participant Checkpointer
    participant DB as Thread database
    Caller->>Cleanup: Delete thread
    Cleanup->>Checkpointer: adelete_thread(thread_id)
    alt Checkpoint deletion succeeds
        Cleanup->>DB: Delete and commit thread row
        Cleanup-->>Caller: Complete
    else Checkpoint deletion raises PsycopgError
        Cleanup-->>Cleanup: Log and return
        Note over Cleanup,DB: Thread row and checkpoint data remain
    end
```
</details>

<sub>Reviews (4): Last reviewed commit: ["test(unit): assert delete was attempted ..."](https://github.com/aegra/aegra/commit/2fa203869ddd5966d6a17ddf1dcaf5c46c6693bf) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=51609578)</sub>

**Context used:**

- Knowledge Base — [Execution Engine: Broker, Executors, Leases, and Cleanup](https://app.greptile.com/aegra/-/custom-context/knowledge-base/aegra/aegra/-/docs/execution-engine.md)

### DIFF
--- a/docs/guides/threads-and-state.mdx
+++ b/docs/guides/threads-and-state.mdx
@@ -146,7 +146,8 @@ for thread in threads:
 
 ## Deleting threads
 
-Deleting a thread cancels any active runs and removes all state:
+Deleting a thread cancels any active runs and removes all state, including
+the checkpoint history stored in the graph backend:
 
 ```python
 await client.threads.delete(thread_id)
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -3,7 +3,7 @@
   "info": {
     "title": "Aegra",
     "description": "Production-ready Agent Protocol server",
-    "version": "0.10.2"
+    "version": "0.10.3"
   },
   "paths": {
     "/info": {
@@ -952,7 +952,7 @@
           "Threads"
         ],
         "summary": "Delete Thread",
-        "description": "Delete a thread by its ID.\n\nPermanently removes the thread and its metadata. Any active runs on the\nthread are automatically cancelled before deletion. Checkpoint history\nstored in the graph backend is not affected.",
+        "description": "Delete a thread by its ID.\n\nPermanently removes the thread, its metadata, and its checkpoint history\nstored in the graph backend. Any active runs on the thread are\nautomatically cancelled before deletion.",
         "operationId": "delete_thread_threads__thread_id__delete",
         "parameters": [
           {
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -990,6 +990,16 @@
               }
             }
           },
+          "500": {
+            "description": "Checkpoint cleanup failed; the thread is preserved and the delete can be retried",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/AgentProtocolError"
+                }
+              }
+            }
+          },
           "422": {
             "description": "Validation Error",
             "content": {
--- a/libs/aegra-api/src/aegra_api/api/threads.py
+++ b/libs/aegra-api/src/aegra_api/api/threads.py
@@ -18,6 +18,7 @@ from aegra_api.core.active_runs import active_runs
 from aegra_api.core.auth_deps import auth_dependency, get_current_user
 from aegra_api.core.auth_filters import build_metadata_filter
 from aegra_api.core.auth_handlers import build_auth_context, handle_event
+from aegra_api.core.database import db_manager
 from aegra_api.core.orm import Run as RunORM
 from aegra_api.core.orm import Thread as ThreadORM
 from aegra_api.core.orm import get_session
@@ -829,9 +830,9 @@ async def delete_thread(
 ) -> dict[str, str]:
     """Delete a thread by its ID.
 
-    Permanently removes the thread and its metadata. Any active runs on the
-    thread are automatically cancelled before deletion. Checkpoint history
-    stored in the graph backend is not affected.
+    Permanently removes the thread, its metadata, and its checkpoint history
+    stored in the graph backend. Any active runs on the thread are
+    automatically cancelled before deletion.
     """
     # Authorization check
     ctx = build_auth_context(user, "threads", "delete")
@@ -865,6 +866,10 @@ async def delete_thread(
                 task.cancel()
                 with contextlib.suppress(asyncio.CancelledError, Exception):
                     await task
+
+    # Checkpoints first: if this fails the thread row survives and the client
+    # can retry; the reverse order would orphan LangGraph checkpoint rows.
+    await db_manager.get_checkpointer().adelete_thread(thread_id)
 
     await session.delete(thread)
     await session.commit()
--- a/libs/aegra-api/src/aegra_api/api/threads.py
+++ b/libs/aegra-api/src/aegra_api/api/threads.py
@@ -36,7 +36,7 @@ from aegra_api.models import (
     ThreadUpdate,
     User,
 )
-from aegra_api.models.errors import CONFLICT, NOT_FOUND
+from aegra_api.models.errors import CONFLICT, NOT_FOUND, AgentProtocolError
 from aegra_api.services.streaming_service import streaming_service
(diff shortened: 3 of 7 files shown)


AayushSaini101:
Thanks for the improvements. There are couple of suggestions provided by coderabbit; please PTAL thanks : ) @arturbagramyan1

AayushSaini101 (approved):

arturbagramyan1:
@AayushSaini101  @ibbybuilds  adressed everything thanks!

arturbagramyan1:
@ibbybuilds on Greptile's remaining concern (transient checkpoint failure abandons the only background-cleanup attempt in `run_cleanup.py`):

This is the pre-existing contract of the fire-and-forget cleanup, not something this PR introduces. On main today, the same single attempt already swallows transient `RedisError`/`SQLAlchemyError`/`OSError` — a transient DB blip already left the ephemeral thread row behind permanently. This PR only adds `PsycopgError` to that existing tuple.

Net effect is strictly an improvement over main:

- **Before:** checkpoint rows leaked on *every* deletion, including successful ones.
- **After:** success cleans everything; on failure the thread row survives (checkpoints are deleted before the row commit), so nothing is orphaned — the leftover is a normal, fully deletable thread.

That retryable residue is exactly what Phase 2's TTL sweep (#288) is designed to collect, and it's documented in the PR description as a known limitation.

That said, if you'd rather not wait for Phase 2, I can add a small bounded retry with backoff (e.g. 3 retries at 1s/10s/60s) around `delete_thread_by_id` in the background path — ~15 lines plus unit tests. Just say the word and I'll push it to this PR.

ibbybuilds (approved):
@arturbagramyan1 verified the whole chain locally on top of the live e2e suite: all 5 thread-deletion e2e tests pass against a real server, and the new checkpoint test proves rows in checkpoints/checkpoint_blobs/checkpoint_writes go from nonzero to zero on delete. checkpoints-first ordering is the right call, failure keeps the thread row so the delete stays retryable, and the ownership-gated call in run_cleanup closes the cross-user hole adelete_thread would otherwise have. no version bump needed, we batch releases now. one note for anyone reading later: the new e2e test needs DATABASE_URL exported to reach the checkpoint tables directly, which make e2e-dev already does

PR #510, opened 2026-08-13
fix(api): serialize common object types

### LINKED ISSUE #480: [BUG] Serializer degrades many common types to repr strings
The `GeneralSerializer` fallback is `return str(obj)`, so any object without `model_dump`/`dict`/`_asdict` silently becomes a string instead of structured data. No error is raised, so consumers receive an unparseable value where they expect an object.

PR #466 fixes this for LangGraph `Command` by adding a dedicated branch. `Command` is a dataclass, so the same failure applies to any plain dataclass in graph state, and to several other common types.

Reproduced against `GeneralSerializer` on main:

| Type | Serializes to | Severity |
|---|---|---|
| plain `@dataclass` | `"AgentState(todos=[Todo(content=...))"` | data loss |
| `bytes` | `"bxffxfex00"` | data loss (see #451) |
| `Enum` | `"Color.RED"` (name, not value) | data loss |
| `deque` | `"deque([1, 2])"` | data loss |
| `complex` | `"(1+2j)"` | data loss |
| `Exception` | `"boom"` (type lost) | data loss |
| `datetime` / `date` | `"2026-01-01 00:00:00"` | wrong type, readable |
| `UUID` | `"7d247e59-..."` | wrong type, readable |
| `Decimal` | `"1.5"` | wrong type, readable |
| `Path` | `"/tmp/x"` | wrong type, readable |

The nested case is the dangerous one, because the surrounding payload still looks valid:

```json
{"messages": [], "state": "AgentState(todos=[Todo(content=\u0027look up customer\u0027...", "ts": "2026-01-01 00:00:00"}
```

A user whose graph state is a plain dataclass hits this today with no error.

## Proposed fix

Extend `GeneralSerializer._serialize_object` rather than adding one branch per class:

- generic `dataclasses.is_dataclass(obj)` branch (this makes #466 `Command` branch redundant)
- `bytes` via base64, which also closes #451
- `Enum` to `.value`
- `datetime`/`date` to isoformat, `UUID`/`Decimal`/`Path` to their natural JSON forms
- `deque` to list

Consider making the final fallback loud (log at minimum) so future gaps surface instead of silently shipping repr strings.

## Related

- #466 fixes the `Command` case, merge on its own merits
- #451 is the `bytes` case
- #453 proposed swapping in orjson for the same family of problems. That approach duplicates serializer logic in a second place; extending the one serializer is preferable.

harsh21234i:
## Summary

  Fixes #480

  Improve `GeneralSerializer` so common Python types are preserved as structured, JSON-compatible values instead of
  silently falling back to `str()` representations.

  ## Root Cause

  Objects without `model_dump`, `dict`, or `_asdict` were converted to strings. This caused data loss for dataclasses
  and common types such as bytes, enums, dates, UUIDs, decimals, paths, deques, and exceptions.

  ## Changes

  - Add recursive serialization for all dataclasses.
  - Encode `bytes` using base64.
  - Serialize enums using their values.
  - Serialize dates and datetimes using ISO 8601.
  - Serialize UUIDs, decimals, and paths as strings.
  - Serialize exceptions with their type and message.
  - Serialize deques as lists.
  - Add regression tests for nested and common object types.
  - Bump package versions to `0.10.1`.

  ## Validation

  - 74 serializer and SSE tests passed.
  - Ruff checks passed.
  - Formatting checks passed.
  - `git diff --check` passed.

  ## Notes

  This change updates the central serializer used by both SSE and broker serialization paths.

## Summary by CodeRabbit

* **Improvements**
  * Enhanced serialization for nested models, dataclasses, collections, mapping keys, and common Python types.
  * Improved JSON compatibility for dates, enums, UUIDs, decimals, paths, bytes, and exceptions.
  * Added safeguards against conflicting serialized dictionary keys.

* **Documentation**
  * Expanded streaming guidance for recursive collection and mapping-key serialization.
  * Added guidance for human-in-the-loop `input.requested` events, including payloads and falsy values.

<h3>Greptile Summary</h3>

The PR expands `GeneralSerializer` to recursively normalize common Python objects for SSE and persisted run output.
- Recursively serializes model dumps, dataclasses, mappings, and collection members.
- Adds structured representations for bytes, enums, temporal values, UUIDs, decimals, paths, and exceptions.
- Documents the resulting wire formats and adds regression coverage for nested values and mapping-key collisions.

The PR appears safe to merge.

No blocking failure remains; the prior model-dump recursion issue and documentation/lint concerns are addressed at the current head.

| Filename | Overview |
|----------|----------|
| libs/aegra-api/src/aegra_api/core/serializers/general.py | Adds recursive normalization for model dumps and common Python values, including normalized mapping keys and collision detection. |
| libs/aegra-api/tests/unit/test_core/test_serializers/test_general.py | Adds regression coverage for common values, nested model dumps, recursive collections, and mapping-key collisions. |
| docs/guides/streaming.mdx | Documents the serializer’s structured JSON representations and mapping-key behavior. |

</details>

<sub>Reviews (5): Last reviewed commit: ["fix(api): normalize serializer mapping k..."](https://github.com/aegra/aegra/commit/dceddb23ff6d6ee3503ca4cbc7eee8e61fc4d9e3) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=53083266)</sub>

**Context used:**

- Knowledge Base — [Authenticated Services and Serializers](https://app.greptile.com/aegra/-/custom-context/knowledge-base/aegra/aegra/-/docs/authenticated-services-and-serializers.md)
- Knowledge Base — [Streaming: from LangGraph execution to SSE wire events](https://app.greptile.com/aegra/-/custom-context/knowledge-base/aegra/aegra/-/docs/streaming.md)

### DIFF
--- a/libs/aegra-api/src/aegra_api/core/serializers/general.py
+++ b/libs/aegra-api/src/aegra_api/core/serializers/general.py
@@ -2,7 +2,14 @@
 
 import dataclasses
 import inspect
+from base64 import b64encode
+from collections import deque
+from datetime import date, datetime
+from decimal import Decimal
+from enum import Enum
+from pathlib import PurePath
 from typing import Any
+from uuid import UUID
 
 from aegra_api.core.serializers.base import SerializationError, Serializer
 
@@ -41,19 +48,32 @@ class GeneralSerializer(Serializer):
         # Command (from tools like write_todos) is a dataclass with no model_dump/
         # .dict()/_asdict; it would otherwise hit str() and reach consumers as an
         # unparseable repr. Emit all fields to match orjson's native output on Platform.
-        elif obj.__class__.__name__ == "Command" and dataclasses.is_dataclass(obj):
+        elif dataclasses.is_dataclass(obj):
             return {field.name: self._serialize_object(getattr(obj, field.name)) for field in dataclasses.fields(obj)}
 
         # Handle NamedTuples (like PregelTask) - they have _asdict() method
         elif hasattr(obj, "_asdict") and callable(obj._asdict):
             return {k: self._serialize_object(v) for k, v in obj._asdict().items()}
 
+        # Handle common scalar types that JSON does not encode natively.
+        elif isinstance(obj, bytes):
+            return b64encode(obj).decode("ascii")
+        elif isinstance(obj, Enum):
+            return self._serialize_object(obj.value)
+        elif isinstance(obj, (datetime, date)):
+            return obj.isoformat()
+        elif isinstance(obj, (UUID, Decimal, PurePath)):
+            return str(obj)
+        elif isinstance(obj, Exception):
+            return {"type": obj.__class__.__name__, "message": str(obj)}
+
         # Handle sets and frozensets
         elif isinstance(obj, (set, frozenset)):
             return list(obj)
 
+        # Handle deques like other JSON array containers.
         # Handle tuples and lists recursively
-        elif isinstance(obj, (tuple, list)):
+        elif isinstance(obj, (deque, tuple, list)):
             return [self._serialize_object(item) for item in obj]
 
         # Handle dictionaries recursively
--- a/libs/aegra-api/src/aegra_api/core/serializers/general.py
+++ b/libs/aegra-api/src/aegra_api/core/serializers/general.py
@@ -35,19 +35,18 @@ class GeneralSerializer(Serializer):
 
         # Handle Pydantic v2 models (model_dump method)
         if hasattr(obj, "model_dump") and callable(obj.model_dump):
-            return obj.model_dump()
+            return self._serialize_object(obj.model_dump())
 
         # Handle LangChain objects and Pydantic v1 models (dict method)
         elif hasattr(obj, "dict") and callable(obj.dict):
-            return obj.dict()
+            return self._serialize_object(obj.dict())
 
         # Handle LangGraph Interrupt objects (they don't have .dict() method)
         elif obj.__class__.__name__ == "Interrupt" and hasattr(obj, "value") and hasattr(obj, "id"):
             return {"value": self._serialize_object(obj.value), "id": obj.id}
 
-        # Command (from tools like write_todos) is a dataclass with no model_dump/
-        # .dict()/_asdict; it would otherwise hit str() and reach consumers as an
-        # unparseable repr. Emit all fields to match orjson's native output on Platform.
+        # Dataclasses (including LangGraph Command) have no standard JSON form.
+        # Emit all fields recursively so nested values retain their structure.
         elif dataclasses.is_dataclass(obj):
             return {field.name: self._serialize_object(getattr(obj, field.name)) for field in dataclasses.fields(obj)}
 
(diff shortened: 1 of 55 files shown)


harsh21234i:
hey @ibbybuilds  whenever you get time please can you review and let me know any changes?

libs/aegra-api/src/aegra_api/core/serializers/general.py
```diff
@@ -28,32 +35,43 @@ def _serialize_object(self, obj: Any) -> Any:
...
+        elif isinstance(obj, (datetime, date)):
+            return obj.isoformat()
+        elif isinstance(obj, (UUID, Decimal, PurePath)):
+            return str(obj)
+        elif isinstance(obj, Exception):
+            return {"type": obj.__class__.__name__, "message": str(obj)}
+
         # Handle sets and frozensets
         elif isinstance(obj, (set, frozenset)):
             return list(obj)
```
ibbybuilds: sets skip the per-item recursion, so a set of Path/UUID/datetime values still crashes persistence the same way this pr fixes elsewhere:

```python
GeneralSerializer().serialize({"tags": {Path("/tmp/a")}})
# -> {"tags": [PosixPath("/tmp/a")]}  -> json.dumps TypeError
```

```suggestion
        elif isinstance(obj, (set, frozenset)):
            return [self._serialize_object(item) for item in obj]
```

ibbybuilds (changes requested):
@harsh21234i verified this fixes the exact crash a user just hit in production (pydantic state carrying a Path field blows up the runs-table JSONB bind with `Object of type PosixPath is not JSON serializable`) - reproduced on main, fixed on your branch, and the full suite is green with main merged in (the 3 ci failures are stale-base flakes fixed on main since you branched, a rebase clears them). one gap left inside this change's own scope, inline below. also please drop the version bump when you rebase, we batch releases now and bump at release time

libs/aegra-api/src/aegra_api/core/serializers/general.py
harsh21234i: done

harsh21234i:
@ibbybuilds I’ve addressed the requested changes.

  - Rebasing/stale-base issue is fixed; the PR is now mergeable.
  - Dropped the version bump by aligning with current `main`.
  - Fixed recursive serialization for `set` / `frozenset` members.
  - Added handling for non-JSON-native mapping keys.
  - Added collision detection for serialized mapping keys, including `UUID` vs string and `1` vs `"1"` cases.
  - Added regression tests and updated the streaming serialization docs.

  Validation run locally:
  - `pytest libs/aegra-api/tests/unit/test_core/test_serializers/test_general.py -q` → 52 passed
  - `ruff check` → passed
  - `ruff format --check` → passed

  CodeRabbit and Greptile are green now, with no new actionable comments. Could you please re-review when you get a
  chance?

ibbybuilds (approved):
@harsh21234i verified on the new head: the set recursion fix covers the gap, the key normalization matches json.dumps semantics for native keys and extends them for uuid/path/enum keys, and a collision degrades to the safe fallback marker instead of a 500. reran the production repro (pydantic state with Path fields) plus sets and mapping keys, all serialize clean, full suite 1886 passed locally. thanks for the quick turnaround and for dropping the bump

PR #603, opened 2026-09-13
fix(api): merge thread metadata in one statement so concurrent writes cannot clobber each other

benlangfeld:
`PATCH /threads/{thread_id}` merges metadata by reading the column into Python, updating the dict and writing the whole document back:

```python
thread = await session.scalar(stmt)
...
if request.metadata:
    current_metadata = dict(thread.metadata_json or {})
    current_metadata.update(request.metadata)
    thread.metadata_json = current_metadata
await session.commit()
```

Nothing locks the row between that read and that write, and the write replaces the whole JSONB document, so two writers that read the same base silently discard each other's keys — a lost update whose window is a whole request round trip.

## Aegra races against itself

This is not only a client-versus-client problem. Run preparation performs the same read-modify-write on the same column (`services/run_preparation.py::update_thread_metadata`), so a client that writes thread metadata while a run is starting on that thread can lose its own keys, or Aegra's. Losing Aegra's is the worse half: `assistant_id` and `graph_id` are written there, and `POST /threads/{id}/state` refuses a thread whose metadata has no `graph_id`, so a normal `PATCH` at the wrong moment can leave a thread unable to accept state updates.

To observe it without patching anything, open two sessions against the same thread, let both read it, then let them commit in turn — the second commit wins whole. The two concurrency tests added here do exactly that, driving the real handlers on separate sessions; both fail on `main` and pass on this branch.

## What changed

Both sites now merge in a single statement, so Postgres reads the column it is writing while holding the row lock:

```sql
UPDATE thread
SET metadata_json = CASE WHEN jsonb_typeof(metadata_json) = 'object'
                         THEN metadata_json ELSE '{}'::jsonb END || :patch
WHERE ...
```

built by a small helper in `utils/jsonb.py` (`jsonb_patch` / `jsonb_shallow_merge`). Two details are load-bearing, both learned from running this in production:

- The patch binds through `JsonbSafe`, the type the JSONB columns already use. Postgres rejects U+0000 in `jsonb`, and a plain `cast(..., JSONB)` would start failing writes that the current Python path accepts, since the Python path never sends the value through the column type at all.
- The `jsonb_typeof` guard is there because `'[1,2]'::jsonb || '{"a":1}'::jsonb` appends to an array instead of merging. Thread metadata is written unvalidated on several paths, and a column holding a non-object now starts from `{}` — which is what `dict(value or {})` meant in Python, except that it raised instead of starting over.

`update_thread` keeps its pre-read: it is the 404 probe and the ownership check, and it no longer feeds the write. In `update_thread_metadata` the "name the thread from the first human message, but only if it has no name yet" rule moves into the statement as a `CASE`, so it too is decided at write time rather than from a read that may already be stale.

## Semantics

Unchanged. `||` is a top-level merge, which is exactly what `update_thread`'s docstring already promised ("shallow merge") and what `dict.update` did: a supplied key replaces that key wholesale, nested objects are not deep-merged, absent keys are untouched. `updated_at` is still bumped for a metadata-free `PATCH`, 404 and ownership behaviour are untouched, and `update_thread_metadata` still does not commit — the merge runs inside the caller's transaction.

The stale "dialect agnostic" note on `update_thread_metadata` is dropped: the column is `JSONB` and the driver is asyncpg, so the read-modify-write was not buying portability.

## Verification

Run on a local Postgres 18:

- `libs/aegra-api/tests/integration/test_thread_metadata_merge_db.py` — six new tests, including the two interleaved-session ones (PATCH versus PATCH, and PATCH versus a starting run), the NUL-byte regression guard, the non-object guard, and the shallow-merge and thread-naming semantics. All pass here; with `api/threads.py` and `services/run_preparation.py` reverted to `main`, the two concurrency tests and the non-object test fail, and the semantics tests still pass.
- `uv run --package aegra-api pytest libs/aegra-api/tests/unit libs/aegra-api/tests/integration` — 2087 passed.
- `make format`, `make lint` clean; `make type-check` reports the same 57 pre-existing diagnostics as `main`, none in the changed files.

The E2E suite was not run here (it needs the compose stack).

One thing worth a maintainer's call: like the existing `tests/integration/test_assistant_large_config_db.py`, these tests skip when no database is reachable, and the unit+integration CI job does not start one — so they will skip in CI as things stand. Happy to add a `postgres` service to that job in this PR if you want them actually running; I left CI alone to keep the change focused.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

<h2><a href="https://app.greptile.com/api/retrigger?id=66438217"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/RetriggerDark.svg?v=2"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/Retrigger.svg?v=2">Retrigger</picture></a>Confidence Score: 4/5</h2>

The implementation appears behaviorally sound, but the repository’s regression-testing requirement must be satisfied by running the new concurrency tests against PostgreSQL in CI before merging.

<h2><a href="https://app.greptile.com/ide/claude-code?prompt=%23%23%23%20Issue%201%0Alibs%2Faegra-api%2Ftests%2Fintegration%2Ftest_thread_metadata_merge_db.py%3A72-81%0AThis%20fixture%20skips%20all%20six%20tests%20when%20PostgreSQL%20or%20the%20migrated%20%60thread%60%20table%20is%20unavailable.%20The%20CI%20job%20that%20collects%20integration%20tests%20does%20not%20start%20PostgreSQL%20or%20run%20migrations%2C%20while%20the%20PostgreSQL-enabled%20E2E%20jobs%20do%20not%20collect%20this%20file.%20As%20a%20result%2C%20the%20two%20tests%20covering%20the%20fixed%20lost-update%20race%20always%20skip%20in%20CI%2C%20and%20the%20unit%20tests%20only%20verify%20the%20generated%20SQL.%20This%20violates%20the%20repository%20requirement%20that%20bug%20fixes%20have%20effective%20regression%20tests%2C%20so%20these%20tests%20must%20run%20against%20a%20migrated%20PostgreSQL%20service%20in%20CI%20before%20merging.%0A%0A---%0A%0AFor%20each%20issue%20above%2C%20determine%20whether%20it%20is%20valid%20and%20should%20be%20fixed.%20If%20so%2C%20fix%20it%20directly.&repo=aegra%2Faegra&pr=603&platform=github"><picture><source media="(prefers-color-scheme: dark)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaudeDark.svg?v=7"><source media="(prefers-color-scheme: light)" srcset="https://greptile-static-assets.s3.amazonaws.com/badges/FixAllInClaude.svg?v=7">Fix All in Claude Code</picture></a>Findings</h2>

1. P2&nbsp;**Concurrency tests skip in CI** <a href="https://github.com/aegra/aegra/pull/603#discussion_r4053783074">▶</a>

<h3>Summary</h3>

This PR replaces Python-side thread metadata read-modify-write operations with atomic PostgreSQL JSONB merges, preserving independently written keys during concurrent PATCH and run-preparation transactions.

- Adds reusable typed JSONB patch and shallow-merge helpers.
- Preserves ownership filtering, 404 behavior, shallow replacement semantics, NUL-byte sanitization, and conditional thread naming.
- Adds SQL-shape, route, and real-PostgreSQL regression coverage, although the database-backed tests are not currently exercised by CI.

<sub>Reviews (1) · Last reviewed commit: ["fix(api): merge thread metadata in one s..."](https://github.com/aegra/aegra/commit/1fdbc7ba6eb49f94fc809f4e6af4d0a6026f2b61)</sub>

### DIFF
--- a/libs/aegra-api/src/aegra_api/api/threads.py
+++ b/libs/aegra-api/src/aegra_api/api/threads.py
@@ -10,7 +10,7 @@ from uuid import uuid4
 
 import structlog
 from fastapi import APIRouter, Depends, HTTPException, Query
-from sqlalchemy import select
+from sqlalchemy import select, update
 from sqlalchemy.dialects.postgresql import insert as pg_insert
 from sqlalchemy.ext.asyncio import AsyncSession
 
@@ -44,6 +44,7 @@ from aegra_api.models.search_limit import effective_search_limit
 from aegra_api.services.streaming_service import streaming_service
 from aegra_api.services.thread_state_service import ThreadStateService
 from aegra_api.services.thread_ttl import get_thread_ttl_config, prune_expired_threads_for_user
+from aegra_api.utils.jsonb import jsonb_patch, jsonb_shallow_merge
 from aegra_api.utils.run_utils import strip_pinned_config_keys
 
 router = APIRouter(tags=["Threads"], dependencies=auth_dependency)
@@ -355,18 +356,27 @@ async def update_thread(
         if isinstance(handler_meta, dict):
             request.metadata = {**(request.metadata or {}), **handler_meta}
 
+    # This read is the 404 probe and the ownership check; the merge below does
+    # not read the column into Python, so it cannot lose a concurrent writer's
+    # keys no matter how long this request takes.
     stmt = select(ThreadORM).where(ThreadORM.thread_id == thread_id, ThreadORM.user_id == user.identity)
     thread = await session.scalar(stmt)
 
     if not thread:
         raise HTTPException(404, f"Thread '{thread_id}' not found")
 
-    thread.updated_at = datetime.now(UTC)
-
+    values: dict[str, Any] = {"updated_at": datetime.now(UTC)}
     if request.metadata:
-        current_metadata = dict(thread.metadata_json or {})
-        current_metadata.update(request.metadata)
-        thread.metadata_json = current_metadata
+        values["metadata_json"] = jsonb_shallow_merge(
+            ThreadORM.metadata_json, jsonb_patch(request.metadata, "metadata_patch")
+        )
+
+    await session.execute(
+        update(ThreadORM)
+        .where(ThreadORM.thread_id == thread_id, ThreadORM.user_id == user.identity)
+        .values(**values)
+        .execution_options(synchronize_session=False)
+    )
 
     await session.commit()
     await session.refresh(thread)
--- a/libs/aegra-api/src/aegra_api/services/run_preparation.py
+++ b/libs/aegra-api/src/aegra_api/services/run_preparation.py
@@ -12,7 +12,7 @@ from uuid import uuid4
 import structlog
 from asgi_correlation_id import correlation_id
 from fastapi import HTTPException
-from sqlalchemy import or_, select, update
+from sqlalchemy import ColumnElement, case, func, literal_column, or_, select, update
 from sqlalchemy.ext.asyncio import AsyncSession
 
 from aegra_api.core.orm import Assistant as AssistantORM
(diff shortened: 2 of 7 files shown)


ibbybuilds (approved):
LGTM

